PR #4, opened 2016-11-29
Provide feedback earlier when mapState is undefined

### LINKED ISSUE #2: Poor feedback if mapState returns undefined
While making the examples for the repo, I found I'd often make a mistake and map the state to undefined by accident (e.g. map to non-existent property of state).  This was difficult to track down as the error is raised by a connected component within the sub-component, or by a generic `undefined` error.

I think there needs to be a check on the result of the `mapState` prop to ensure it is not `undefined` before proceeding.

I would also suggest that the check should only be done if running in a non production environment as this is more of an aid for developers when setting up the state mappings than something we'd want users to see in their consoles.

mpeyper:
This resolves #2

### DIFF
--- a/lib/utils/subState.js
+++ b/lib/utils/subState.js
@@ -21,7 +21,11 @@ var getSubState = exports.getSubState = function getSubState(getState, mapState)
         var rootState = getState();
         var subState = mapState(rootState);
 
-        if (subState && (typeof subState === 'undefined' ? 'undefined' : _typeof(subState)) === 'object' && !Array.isArray(subState)) {
+        if (process.env.NODE_ENV !== 'production' && subState === undefined) {
+            throw Error('mapState must not return undefined.');
+        }
+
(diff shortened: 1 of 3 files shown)


jpeyper:
should we use [console.assert](https://developer.mozilla.org/en-US/docs/Web/API/console/assert) instead?

mpeyper:
Possibly.  Something that concerns me is a note in the docs: 

> Specifically, in browsers, calling console.assert() with a falsy assertion will cause the message to be printed to the console without interrupting execution of subsequent code.

Throwing an error prevents the component from rendering at all, making it really obvious something has gone wrong.

mpeyper:
So I played around with console.assert and I actually prefer the result.  

The component still doesn't render as the mapStateToProps throws an undefined error and prevents the render from occurring.  However, when looking in the console, you see an explanation of why the undefined might have occurred.  

The other big improvement is that you are also told which component is being incorrectly mapped.  the previous attempt could only tell you that one of subspaced components was not right, but it doesn't actually know which one.

I have updated the PR with console.assert.

PR #10, opened 2016-12-06
Add TravisCI build config

chaos95:
Basic config to build and test the project against nodejs 6.x.

Runs `npm install` and `npm test` as per TravisCI defaults.


mpeyper:
thanks :)

PR #21, opened 2016-12-21
Increased Test Coverage

### LINKED ISSUE #14: Unit Test SubspaceProvider
I would like to have a unit test around the [`SubspaceProvider` component](https://github.com/ioof-holdings/redux-subspace/blob/master/src/components/SubspaceProvider.js) but I don't have much experience testing react components, especially those that use features like context and Redux.

Any assistance (either pointers or pull requests) on doing this would be much appreciated.

mpeyper:
Added test for SubspaceProvider
Added code coverage
Cleaned up names of dispatch tests
Fixes #14

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 node_modules
 npm-debug.log
+.nyc_output
+coverage
 .vscode
+.idea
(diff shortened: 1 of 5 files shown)


mpeyper:
I'm just going to merge this as there are no functional changes and its been sitting here a while, so if anyone had any objections they would have raised them by now.

PR #25, opened 2017-01-18
Added global action registration and as global helper function

### LINKED ISSUE #24: Suggestion to improve global actions from redux middlewares/addons
In using subspace, I discussed a few issues with using [react-router](https://github.com/ReactTraining/react-router) in #22 

One of the other issues I have had is using [react-router-redux](https://github.com/reactjs/react-router-redux) and the actions it raises to do navigation.

These need to be global actions, so I have something like this...

```javascript
import { push as realPush, goBack as realGoBack } from 'react-router-redux'

export const push = path => {
    return {
        globalAction: true,
        ...realPush(path)
    }
}

export const goBack = path => {
    return {
        globalAction: true,
        ...realGoBack(path)
    }
}
```

react-router-redux exports its action type [`LOCATION_CHANGE`](https://github.com/reactjs/react-router-redux/blob/master/src/index.js).

It would be nice to be able to register this action type against subspace as a global action so that I don't have to wrap all its actions.

I'm sure there are other action types that always need to be global to work properly.

Also, requiring to know the `globalAction: true` seems unnecessary and limiting for changing how global actions may work in the future. Adding an `asGlobalAction(action)` function would clean up the implementation.

mpeyper:
resolves #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,13 +121,33 @@ const reducer = combineReducers({ subComponent: namespaced(subComponent, 'myComp
 
 #### Global Actions
 
-Occasionally you may have actions that need to go beyond your small view of the world.  Adding `globalAction: true` to your action will tell the `SubspaceProvider` to not prepend the namespace and `namespaced` to not exclude it from the reducer.
+Occasionally you may have actions that need to go beyond your small view of the world.  
+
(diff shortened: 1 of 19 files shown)


vivian-farrell:
This is a much better way than relying on hardcoding 'globalAction'.

PR #26, opened 2017-05-16
Provide root state as second parameter to mapState

mpeyper:
Having to constantly check if `state.root` exists is annoying and reduces our ability to refactor the `root` node out in the future.  Providing the `rootState` to the mapState function simplifies both.

I have also included TypeScript definitions for subspace in this PR.


vivian-farrell:
Looks good to me

Issue #18, opened 2016-12-15
Limited compatibility with redux-sagas


mpeyper:
As identified in #8, there is only partial compatibility with [redux-sagas](https://github.com/yelouafi/redux-saga).

There appear to be no issues until the [namespacing](https://github.com/ioof-holdings/redux-subspace#namespacing) feature is used.  The actions being dispatched by sagas are not passing through our wrapper and the namespaces are not being applied so the reducers are discarding them.

I have taken a look at the problem myself, but without much exposure to sagas before, I'm struggling to find a solution so any help would be much appreciated.

amitkothari:
Hi Michael, 

I have added a simple redux-saga example with namespacing feature. 
[https://github.com/amitkothari/redux-subspace](url)

mpeyper:
Hey Amit,

Thanks for contributing!

I think the files I want to be concentrating on are the [component sagas](https://github.com/amitkothari/redux-subspace/blob/master/examples/sagas/component/sagas.js) and the [app sagas](https://github.com/amitkothari/redux-subspace/blob/master/examples/sagas/app/sagas.js), right?

Firstly, when I'm looking at this, I'm keeping subspace's [key concepts](https://github.com/ioof-holdings/redux-subspace/blob/master/CONTRIBUTING.md#key-concepts) in mind, so having the parent control the bounds of the subspace and the child being unaware of the parent 

Ideally, the interface when integrating the component sagas would look something like:

```
import { sagas as componentSagas } from '../component'
import { all } from 'redux-saga/effects'
import { namespacedSagas } from 'redux-subspace'

export default function* sagas() {
  yield all([ 
    namespacedSagas(componentSagas, 'component1') , 
    namespacedSagas(componentSagas, 'component2') 
  ])
}
```

where the component exports unmodified sagas and subspace takes care of the rest.  From my understanding (which is admittedly limited on sagas), the issue here is that we don't have a way of intercepting the actions before the `takeEvery` and after the `put` because they are imported and called directly (i.e. not passed in like `dispatch` is to a thunk).

Because of this, it may be necessary to bleed some of the subspace logic into the component.  When doing so though, it should be in a way that the parent (app in this case) is still in control and so that the component has to know as little about subspace's implementation as possible.  For example, if the app sagas did not pass through a namespace, you example would not work.

If the component sagas looked something like:

```
import { takeEvery } from 'redux-saga'
import { put } from 'redux-saga/effects'
[...]

Issue #31, opened 2017-07-08
Add optional scope parameter when namespacing


mpeyper:
This is going to get a bit confusing to explain, but I'll do my best.

I've run into an issue where I want to add a `namespaced` reducer for a `subspaced` component, but the reducer will be a sibling of parent component's reducer, instead of a child.

Component structure:
- App
  - Parent (namespace: `'parent'`)
    - Child (namespace: `'child'`)

Reducer structure:
- Root
  - Parent (namspace: 'parent')
  - Child (namespace: 'child')

The reason why the reducer is a sibling and not being combined into the parent's reducer is because the reducers are being added dynamically using the [`replaceReducer'](http://redux.js.org/docs/api/Store.html#replaceReducer) which, so far, I have only been able to get reducers to be added to the top level of the store.

The problem is, when an action (e.g. `'SET_VALUE'`) is dispatched from the child, as it breaks out of each subspace, the namespace gets prepended so the final action type is `'parent/child/SET_VALUE'` but the action is ignore by both reducers:
- The parent reducer will successfully remove the `'parent'` namespace but none of it's reducers are listening for the resulting action type (`'child/SET_VALUE'`).
- The child reducer is looking for an action that starts the `'child\'` so it won't pass on an action that starts with `'parent/child'`.

I propose the namespacing functionality be expanded to include an `scope` parameter to apply the namespace to.  If passed `undefined`, if will fallback to the default behaviour of inferring the scope from the nesting hierarchy.  I also propose that the same functionality be available anywhere a namespace can be provided (`namspace`, `subspaced`, `SubspaceProvider`)

It could be implemented as:
1. a seperate optional parameter
    - `namespace(reducer, 'child', 'parent')`
    - `subspaced(mapState, 'child', 'parent')(Component)`
    - `<SubspaceProvider mapState={mapState} namespace='child' scope='parent'>`
2. optionally accepting an object for the `namespace parameter
    - `namespace(reducer, { id: 'child', scope: 'parent' })`
    - `subspaced(mapState, { id: 'child', scope: 'parent' })(Component)`
    - `<SubspaceProvider mapState={mapState} namespace={{ id: 'child', scope: 'parent' }}>`

Another thing to consider what are acceptable values for `scope`.  I think it should accept:
1. a `string`
    - most basic way to add a scope
    - `namespace(reducer, { id: 'child', scope: 'parent' })` would result in the reducer being `namespaced` to `'parent/child'`
2. an `array`
[...]

Issue #19, opened 2016-12-15
Improved documentation for supported Redux middleware


mpeyper:
It was suggested in #8 that we could improve the documentation around support, limitations and incompatibility of Redux middleware.

As it stands, [redux-thunk](https://github.com/gaearon/redux-thunk) is the only supported middleware and there is a known limitation with [redux-sagas](https://github.com/yelouafi/redux-saga) (#18).

Any suggestions on the format of documentation (or a pull request) are welcome.

mpeyper:
Closing this as there is now [a section in the docs](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/README.md) for middleware and improved middleware support introduced in #34.

PR #34, opened 2017-08-07
redux-subspace version 2 - Spliting packages, better middleware support, redux-saga support, improved docs

mpeyper:
This one is huge.  I don't think a single line of the version 1 survived.  It will probably be easier to just look at or clone [my fork](https://github.com/mpeyper/redux-subspace) than to read this PR, but review however you feel is easiest for you.

The changes here are plentiful:

- Complete refactor of core subspace code
  - Middleware support
    - Improved support for Redux middleware
    - Added middleware pipelines for subspaces as well
  - More versatile ways to create subspaces (less duplication in keys for common cases)
- Library has been split into multiple packages
  - redux-subspace: The core of the library
  - react-redux-subspace: React bindings to create subspaces for components
  - redux-subspace-saga: Support for creating subspaced sagas compatible with redux-saga
  - redux-subspace-wormhole: A redux-subspace middleware that replaces the `root` node functionality
- Improved documentation


mpeyper:
I haven't done much (read: any) spell/grammar checking on the new docs, but I really wanted to get this up for review.  If you see any, comment them an I'll fix them up in due course.

mpeyper:
Also, this should resolve #18, #19, #31.

mpeyper:
You can [see the build here](https://travis-ci.org/mpeyper/redux-subspace) with all the tests that run for each package.

vivian-farrell:
Really great effort Michael. I will start digging through.

mpeyper:
Just added a small change to the saga middleware.  Now, when using the replacement `createSagaMiddleware` you don't need to wrap the root saga in `provideStore` (it handles this automatically now).

vivian-farrell:
Is the import in https://github.com/mpeyper/redux-subspace/blob/master/packages/react-redux-subspace/src/components/SubspaceProvider.js for subspace correct? Redux-subspace is not in the list of packages in the package.json so how is it resolved?

vivian-farrell:
My bad, it is in the list. Ignore the comment above.

jpeyper:
The code looks good and test coverage is awesome. Good job.

Docs also seem good, although I think there may be some tweaking over time.

I think we should merge this as is and then fix any docs problems on master.

PR #41, opened 2017-08-24
Bug Fix: Issue when middleware used middleware

mpeyper:
Basically, if a middleware tried to use middleware (e.g. [thunk dispatching a thunk](https://github.com/ioof-holdings/redux-subspace/pull/41/files#diff-49119f9350fb9b82968e43f8740e434c)), the second middleware would use the parent subspace's scope, not the originating subspace.

This PR fixes that.

### DIFF
--- a/packages/redux-subspace/src/enhancers/applySubspaceMiddleware.js
+++ b/packages/redux-subspace/src/enhancers/applySubspaceMiddleware.js
@@ -12,14 +12,28 @@ const applySubspaceMiddleware = (...middlewares) => (createSubspace) => (store) 
 
     const subspacedStore = createSubspace(store)
 
-    const chain = middlewares.map((middleware) => middleware(subspacedStore))
+    let getState = subspacedStore.getState
+    let dispatch = subspacedStore.dispatch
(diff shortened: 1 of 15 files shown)


packages/redux-subspace/src/enhancers/applySubspaceMiddleware.js
```diff
@@ -12,14 +12,28 @@ const applySubspaceMiddleware = (...middlewares) => (createSubspace) => (store)
...
 
-    const chain = middlewares.map((middleware) => middleware(subspacedStore))
+    let getState = subspacedStore.getState
+    let dispatch = subspacedStore.dispatch
+
+    const { namespace, rootStore, subspaceTypes, processAction } = subspacedStore
+
+    const middlewareApi = {
+        getState: (...args) => getState(...args),
+        dispatch: (...args) => dispatch(...args),
```
mpeyper: This is now much more similar to [Redux's own implementation](https://github.com/reactjs/redux/blob/master/src/applyMiddleware.js#L27).

Issue #38, opened 2017-08-21
Middleware Compatibility


mpeyper:
After the v2 improvements in #34 we are in a better position to support more existing redux middlewares.

This started with [redux-saga](https://redux-saga.js.org/) (which was added in #34), then [redux-promise](https://github.com/acdlite/redux-promise) (#36) and now [redux-observable](https://redux-observable.js.org/) (if #37 is merged).

My plan is to try out the most popular middleware and either document that they work, or list any caveats in the [redux-subspace documentation](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/README.md) (like [redux-thunk](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/redux-thunk.md) and [redux-promise](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/redux-promise.md)), and, if required create a compatibility package for them (like [redux-saga](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/redux-saga.md)).

To start with, I want to at least cover the middleware in the [ecosystem section of the Redux docs](http://redux.js.org/docs/introduction/Ecosystem.html#middleware), but also other's as they come on my radar, such as [redux-logic](https://github.com/jeffbski/redux-logic) which was requested in #30.

This would be a great task for others to get their hands dirty in redux-subspace.

mpeyper:
## Middleware:

- [x] [redux-thunk](http://github.com/gaearon/redux-thunk) - [documented](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/redux-thunk.md)
- [x] [redux-promise](https://github.com/acdlite/redux-promise) - [documented](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/redux-promise.md)
- [] [redux-axios-middleware](https://github.com/svrcekmichal/redux-axios-middleware)
- [] [redux-observable](https://github.com/redux-observable/redux-observable/) - [compatability package PR](https://github.com/ioof-holdings/redux-subspace/pull/37)
- [] [redux-cycles](https://github.com/cyclejs-community/redux-cycles)
- [] [redux-logger](https://github.com/fcomb/redux-logger)
- [] [redux-immutable-state-invariant](https://github.com/leoasis/redux-immutable-state-invariant)
[...]

PR #44, opened 2017-08-27
redux-loop compatibility

rikukissa:
[redux-loop](https://github.com/redux-loop/redux-loop) is quite popular library being used in conjunction with Redux. It seems to be working nicely with redux-subspace, except for the fact that the side effects returned by the reducers do not get namespaced like they should.

Basically this library would be a wrapper around the default `namespaced` higher-order reducer that takes in the "loop" returned by the reducer and modifies the command so that all the actions included will be namespaced.

This PR is still WIP and is lacking documentation

Comments are welcome :)

https://github.com/ioof-holdings/redux-subspace/issues/38


jpeyper:
Thanks for your PR.

We will take a look at this (and redux-loop) over the next few days :)

One question, you mentioned compatiblility with redux-loop v2. Do you feel this is important? Do you think it will be difficult to have compatibility with both within the same package?

My personal thoughts is to support the latest version only.

rikukissa:
I would go with supporting only the latest version as well, but unfortunately I need the v2 support myself 😔 The API changes they've made between v2 and v3 have been quite drastic and since I'm currently working on a fairly large codebase, upgrading isn't currently an option.  

Support for the previous version is easy to implement and I'd argue it would be a good thing for this library, especially from the easy adoption point of view. I'm guessing there's a lot of projects still stuck with v2.

rikukissa:
I have to take my word back about it being easy to support both versions at the same time. Turned out to be much more difficult than I thought. Lets only support the latest version like you suggested. I'll do fork for v2.0 at some point.

mpeyper:
Can you branch the logic by the available redux-loop functions?

Something like

```
import * as ReduxLoop from 'redux-loop'

if (ReduxLoop.getCmd && ReduxLoop.getEffect) {
	// version 3
	// handle either
} else if (ReduxLoop.getEffect) {
	// version 2
	// only handle effects
} else {
	// version 4 (or whenever effects get removed)
	// only handle commands
}
```

As for publishing a forked version, you could try something like [what redux-loop suggests](https://github.com/redux-loop/redux-loop#dont-see-a-feature-you-want).

rikukissa:
Tried doing that, and it would actually work. The problem that I had was related to testing whether our logic is correct for handling the effects created by redux-loop v2. In our unit tests I would have to manually create the v2 style effects, since redux-loop's v3 API doesn't provide a way for generating them. They do have a module called [effects.js](https://github.com/redux-loop/redux-loop/blob/master/modules/effects.js) for backwards compatibility reasons, but it's basically a wrapper around the new command system. 

Now the problem with creating them manually is that our [isLoop call](https://github.com/ioof-holdings/redux-subspace/pull/44/files#diff-6d98d47b7281e2611bec439e9fff2673R49) starts returning undefined because the [logic behind it](https://github.com/redux-loop/redux-loop/blob/master/modules/cmd.js#L19-L21) relies on a symbol that isn't exported from redux-loop at all.

So the options I see here are
  - Implement the v2 support without unit tests
  - Publish v2 version as a separate package. Here we could install v2 as a devdep and use it in tests

Thanks for the link 👍  That would work indeed. Didn't even think of that to be honest.

rikukissa:
Hey I think I figured out another solution! Maybe we could first create v3 style commands and modify them to look like v2 effects while still keeping the symbol key in the effect object. I'll try that and report back!

rikukissa:
Hey, it worked! The only downsides of doing this is that we now get some warnings to our test output: 
image

and that I needed to create [this factory function](https://github.com/ioof-holdings/redux-subspace/pull/44/files#diff-6d98d47b7281e2611bec439e9fff2673R78) so I'm able to mimic the v2 environment in our tests.

Tried it in 2 of my projects running different version of redux-loop and it seems to be working fine 🙂

mpeyper:
Great!  We can look at suppressing console output for those test cases, but I don't think it's necessary for this PR atm.

packages/redux-subspace-loop/package.json
```diff
@@ -0,0 +1,61 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ioof-holdings/redux-subspace.git"
+  },
+  "dependencies": {
+    "redux-subspace": "^2.0.3-alpha"
+  },
+  "peerDependencies": {
+    "redux-loop": ">=2.0.0 <3.0.0"
```
mpeyper: should this be `^2.0.0 || ^3.0.0`?

packages/redux-subspace-loop/test/integration-spec.js
```diff
@@ -0,0 +1,111 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { createStore } from 'redux'
+import { Cmd, loop, install, combineReducers } from 'redux-loop'
+import { namespaced } from '../src'
+import { subspace } from 'redux-subspace'
+
+describe('integration tests', () => {
```
mpeyper: I would like to see `should work with namespaced single subspace` and `should work with namespaced nested subspaces` integration tests as well.

packages/redux-subspace-loop/package.json
rikukissa: Good point! Fixed.

packages/redux-subspace-loop/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        if (effect.type === 'BATCH') {
+            return {
+                ...effect,
+                effects: effect.effects.map(namespaceEffect)
+            }
+        }
+
+        return effect
```
mpeyper: code coverage shows this line is uncovered (not super important, just FYI).

packages/redux-subspace-loop/docs/Usage.md
```diff
@@ -0,0 +1,26 @@
+# Usage
+
+Using Redux Subspace with [`redux-loop`](https://github.com/redux-loop/redux-loop) is pretty straightforward and should require no additional enhancer or middleware configuration. The only difference is that you have to import the `namespaced` function from `redux-subspace-loop` instead using the default one redux-subspace package. This extends the functionality provided by redux-subspace to also namespace the commands returned by your reducer.
```
mpeyper: > The only difference is that you have to import the `namespaced` function from `redux-subspace-loop` instead **_of_** using the default one **_from_** the redux-subspace package

Just a suggestion, feel free to reword your own way.

packages/redux-subspace-loop/docs/api/namespaced.md
```diff
@@ -0,0 +1,20 @@
+# `namespaced(namespace)`
+
+A higher-order reducer that filters out actions that do not match the provided namespace and namespace the commands returned by your reducer.
```
mpeyper: > A higher-order reducer that filters out actions that do not match the provided namespace and **_namespaces_** the commands returned by your reducer.

Again, just a suggestion

mpeyper:
I've looked through the code and the approach, naming and result is consistent with redux-subspace and the other compatibility packages.  Good job!

There should probably be a page added in the main docs similar to [the middleware page](https://github.com/ioof-holdings/redux-subspace/blob/master/docs/advanced/middleware/README.md) about compatible enhancers.  I'm not asking you to do this (unless you want to), but it should be done shortly after this is merged.

mpeyper:
@rikukissa Let me know if you want to make the suggested changes yourself before I merge this, or if you want me to merge it and someone (probably me) can make them later.

Either way, I'm happy for this to be merged.

packages/redux-subspace-loop/test/integration-spec.js
rikukissa: Will add them today :)

rikukissa:
Thanks for the review! I'll make some changes tonight so we can get this merged 👍

rikukissa:
Thanks again for the review @mpeyper!  Addressed all the things you mentioned and from my perspective, this PR should now be good to go. Might make sense to take a quick look at the commits I pushed tho 🙂

rikukissa:
Suppressed the redux-loop effect nags :)

image

mpeyper:
Looks good.

PR #45, opened 2017-08-29
Fix license badge links in packages

jpeyper:
I noticed that if you clicked the license badges they were not pathing to the license at the root

### DIFF
--- a/packages/redux-subspace/README.md
+++ b/packages/redux-subspace/README.md
@@ -2,7 +2,7 @@
 
 [![npm version](https://img.shields.io/npm/v/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
 [![npm downloads](https://img.shields.io/npm/dm/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
-[![License: MIT](https://img.shields.io/npm/l/redux-subspace.svg?style=flat-square)](../../LICENSE)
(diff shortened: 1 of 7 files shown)


jpeyper:
@vivian-farrell @mpeyper when we publish the npm packages, should the license be copied in?

packages/redux-subspace/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![npm version](https://img.shields.io/npm/v/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
 [![npm downloads](https://img.shields.io/npm/dm/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
-[![License: MIT](https://img.shields.io/npm/l/redux-subspace.svg?style=flat-square)](LICENSE)
+[![License: MIT](https://img.shields.io/npm/l/redux-subspace.svg?style=flat-square)](/LICENSE.md)
```
jpeyper: this was just wrong before.

mpeyper:
I'm happy with this.  @vivian-farrell will be best to answer if the license file needs to be in the published package or if the license field in the package.json and the the link to the one in the root of the repo is enough.

jpeyper:
Lets merge this now, and if @vivian-farrell gets back to us I'll do another fix to do whatever is required.

mpeyper:
I guess it less incorrect now than before.  Still might be incorrect.

PR #48, opened 2017-09-05
Logos

thatalextaylor:
Logos in 512x512, 128x128 and 32x32. Inkscape SVG and PNG. The 32x32 PNG has been hand edited for clarity.


mpeyper:
Hey... its THAT Alex Taylor!

Those are some [lovely shade's of green](https://www.ioofonline.com.au/images/logo.svg) you've chosen! ;)

My biggest concern is that the logo doesn't have a strong association to [redux](https://raw.githubusercontent.com/reactjs/redux/master/logo/logo.png) or [subspace](https://en.wikipedia.org/wiki/Technology_in_Star_Trek#Subspace) (I was secretly hoping for something with a spaceship in it).  This isn't a deal breaker, just something to consider.

Thanks for the time and effort you put into these.

vivian-farrell:
@mpeyper do you want these? Or do you think we should iterate on the design? I think the concept is good but I would try to make it a bit simpler. Great work @thatalextaylor

mpeyper:
@vivian-farrell I have spoken to @thatalextaylor about them and the decision was not to go forward with these ones.  The takeaway was that we wanted something a bit cleaner and minimalistic, and preferably with a tie in to redux (it is a redux extension library after all).

@thatalextaylor are you working on new ones (yet?)?  I'm not sure if we should keep this open as a reminder or if it's better to open new ones for each design?

thatalextaylor:
I'll make a new one for new logos.

PR #50, opened 2017-09-12
Added redux-subspace-loop example

mpeyper:
Added a port of the [official redux-loop example](https://github.com/redux-loop/redux-loop/tree/master/example) that puts multiple instances on the page at the same time.

You can see it working on my fork in [codesandbox](https://codesandbox.io/s/github/mpeyper/redux-subspace/tree/master/packages/redux-subspace-loop/examples/redux-loop-example).

I also made a change to how redux-subspace-loop is exporting so that the `index.js` only has a single export (`namespaced`) as previously the `createNamespacer` function was also being exported (for testing) but was not expected or documented.

Finally, I added a seperate [`Enhancers` section to the docs](https://github.com/mpeyper/redux-subspace/tree/master/docs/advanced/enhancers) to differentiate it from `Middleware`.


mpeyper:
Oh, this also includes change babel-preset-es2015 and babel-preset-stage-0 to babel-preset-env and babel-preset-stage-3 as per a conversation between @vivian-farrell and myself.  See [this](https://babeljs.io/docs/plugins/#presets-official-presets) for details about babel-preset-env.

PR #54, opened 2017-10-27
[react-redux-subspace] Add string to prop types of mapState prop

mradionov:
According to the docs `mapState` may receive a string, but when I use `<SubspaceProvider mapState="someString" />` React throws a warning about invalid prop type.

This PR extends prop types for the `<SubspaceProvider>` component to expect either string or function.

### DIFF
--- a/packages/react-redux-subspace/src/components/SubspaceProvider.js
+++ b/packages/react-redux-subspace/src/components/SubspaceProvider.js
@@ -27,7 +27,10 @@ class SubspaceProvider extends React.PureComponent {
 
 SubspaceProvider.propTypes = {
     children: PropTypes.element.isRequired,
-    mapState: PropTypes.func,
+    mapState: PropTypes.oneOfType([
+        PropTypes.func,
+        PropTypes.string,
+    ]),
     namespace: PropTypes.string,
     subspaceDecorator: PropTypes.func,
 }


mpeyper:
Of course! Good pick up.

I'm away from home until tomorrow night, so I'll merge this and push a new version then, unless on of the others get to it first.

mradionov:
@mpeyper btw, thank you for the great lib. My project got to the point when I had to use same components with their Redux infrastructure (reducer, actions, selectors) multiple times in the project. By default in Redux everything is so tightly coupled and it's pretty difficult do achieve some kind of reuse without code duplication. So I had to start thinking about how to make it work, so I've identified the problems which were stopping me from reuse:
1) reducer unique position in state tree and it's unique name
2) selectors which depended on the reducer position in state tree
3) actions got dispatched for all mounted instances
I've even started working on some kind of proof of concept with higher-order reducers and action namespacing, when I saw your posts in [this React issue](https://github.com/reactjs/react-redux/issues/278). I was very happy to find a library which has everything I was thinking about of doing by myself and it has even more than I needed, thank you for spreading the word. Your implementation is very nice and pure, I am very grateful for your efforts of putting this library together. Good job!

mpeyper:
😊  Thanks @mradionov.  I'm glad you find it useful.

PR #56, opened 2017-12-22
Pass redux-saga context through to subspaced sagas

mpeyper:
redux-saga context is not being passed through to subspaced sagas as they run in a different saga runtime to the root store middleware.

This change will pass on any  context provided during middlware creation so that the subspaced saga has access to it as well.

There is a caveat that context provided by the [`setContext` effect](https://redux-saga.js.org/docs/api/#setcontextprops) will not be preserved due to a limitation in how context is accessed within sagas.  There is a [redux-saga issue](https://github.com/redux-saga/redux-saga/issues/1280) to hopefully get around then, but until then, we are restricted in what we can provide.  I have documented this in the `redux-subspace-saga` usage docs.

### DIFF
--- a/packages/redux-subspace-saga/src/middleware/createSagaMiddleware.js
+++ b/packages/redux-subspace-saga/src/middleware/createSagaMiddleware.js
@@ -13,7 +13,7 @@ export default (options) => {
     const sagaMiddleware = createSagaMiddleware(options)
     
     const sagaSubspaceMiddleware = (store) => {
-        sagaSubspaceMiddleware.run = (saga) => sagaMiddleware.run(provideStore(store)(saga))
+        sagaSubspaceMiddleware.run = (saga, ...args) => sagaMiddleware.run(provideStore(store, options)(saga), ...args)
         return sagaMiddleware(store)
     }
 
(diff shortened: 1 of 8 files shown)


packages/redux-subspace-saga/src/middleware/createSagaMiddleware.js
```diff
@@ -13,7 +13,7 @@ export default (options) => {
     const sagaMiddleware = createSagaMiddleware(options)
     
     const sagaSubspaceMiddleware = (store) => {
-        sagaSubspaceMiddleware.run = (saga) => sagaMiddleware.run(provideStore(store)(saga))
+        sagaSubspaceMiddleware.run = (saga, ...args) => sagaMiddleware.run(provideStore(store, options)(saga), ...args)
```
mpeyper: This isn't directly related to the change, but I noticed that [`run` accepts args](https://redux-saga.js.org/docs/api/#middlewarerunsaga-args) to pass on to the saga, so our wrapper should too.

packages/redux-subspace-saga/src/sagas/subspaced.js
```diff
@@ -39,17 +39,19 @@ const subspaced = (mapState, namespace) => {
     const subspaceDecorator = subspace(mapState, namespace)
 
     return (saga) => {
-        return function* wrappedSaga() {
+        return function* wrappedSaga(...args) {
```
mpeyper: Similar reasons to why we're [passing `args` through in the middleware](https://github.com/ioof-holdings/redux-subspace/pull/56/files#r158407618)

jpeyper:
Looks great.

Issue #60, opened 2018-01-06
Create a redux-subspace logo


mpeyper:
I've [previously stated](https://twitter.com/michaelpeyper/status/904650745298825216) that I would like to get a (cool) logo for this library, but having zero artistic ability myself, I'm reaching out to anyone who does.

There was [one attempt a while ago](https://github.com/ioof-holdings/redux-subspace/pull/48) with some notes in it of what I'm after but here are some general ideas (you don't have to use all of them):

- minimalistic, clean design
- space themed (stars, rockets, planets, moons, etc.)
- tied in to the [Redux logo](https://github.com/reactjs/redux/tree/master/logo) would be nice (e.g. [redux-saga](https://github.com/redux-saga/redux-saga/tree/master/logo)), keeping in mind we integrate with a lot of other libraries of the redux ecosystem as well, so being able to switch out the redux logo for their logo is a win
- could use the ["IOOF greens"](https://www.ioofonline.com.au/images/logo.svg) for colour inspiration
- ignore everything I've said and make something awesome

matthewhartman:
I had a little attempt this morning. I'll send a PR for it :)

mpeyper:
I've given up waiting for something I like.  Closing.

PR #61, opened 2018-01-07
#60 - Added Subspace logo

matthewhartman:
Let me know what you think.

I can do more tweaks to this current version or go down a different approach :)


mpeyper:
My initial reaction was that I really liked it!

I made a few small notes:

- Perhaps the loops could be different colours? IOOF dark green, IOOF light green and Redux purple?
- 2 of the loops are open with 1 dot and the other is closed with 2 dots.  Intentional?
- Make the 'S' a bit bigger so it's more identifiable at favicon size

The more I looked at it, the more familiar it felt... Then it hit me... it's VERY similar to the React logo

image

It's actually a pretty nice mash up of the React and Redux logos and could almost be a good logo for the [`react-redux` package](https://github.com/reactjs/react-redux).

image

image

I was sort of ok with the similarities, because as I said, I quite like the design, but then I stumbled across another popular library's logo when looking for links for this response... [`react-semantic`](https://react.semantic-ui.com/introduction) and it's a bit too close for my liking.

image

image

So I like the direction of this, but I'm not sure what can be done to make it more unique from other packages?

matthewhartman:
Yeah I tried to mash up the react and redux logo into like an 'atom' and each circle would represent a 'store' and it's all connected, so each store is talking to another.

I still think it's quite different to the `react-semantic` package logo (this is the first time I'm seeing this logo).

I can add circles to where those missing gaps are so it's all connected up.

If you're not happy with it still, I can try for a different direction.

Looking forward to hearing your thoughts.

mpeyper:
I think I'd like it less on the React path and more on the Redux path.  By that I mean that redux-subspace is an extension of the redux ecosystem and it's only because their's touches React that we have to too.

> each circle would represent a 'store' and it's all connected, so each store is talking to another

If we're trying to bring in conceptual tie-ins to the logo, then the store interaction is actually more about isolation and preventing cross-talk between the stores.  Composition of subspaces (nesting stores in stores) is another area where a conceptual tie-in for a logo might be possible.

> I still think it's quite different to the `react-semantic` package logo

I really liked it until I saw react-semantic's, and I can't un-see it now 😞 

> Sorry for the late reply, I got a hall pass for a couple of hours ;)

Lucky you!

mpeyper:
> If you're not happy with it

Also, I'd like to stress that this is not just my library.  Anyone from IOOF can review and accept changes.  I'm happy to be outvoted if others want it.

matthewhartman:
If each store is isolated, I'm thinking something more like a universe where you've got multiple planets or stars, which are independent to each other.

Let me have a think about it for some time and come up with another concept.

Cheers.

matthewhartman:
https://github.com/ioof-holdings/redux-subspace/blob/e83d89aa6fe7f678b44b224790f6f59f835c66c1/README.md

Those comets look a little out of whack. The colours might need tweaking too.

matthewhartman:
Check it out mate:
https://github.com/ioof-holdings/redux-subspace/blob/b326f6fc4dd6b838eb9d9fb6677d962ae4511d42/README.md

jpeyper:
That's looking awesome. I love it.

mpeyper:
@matthewhartman and myself have been discussing this over slack (faster feedback than github issues), but I thought I should put a summary here for other to see:

- we like the mix of "redux purple" and "ioof greens"
- we're not sure on the grey-blue "comets"
   - mainly the colour is off, conceptually they are fine

This concept gave me an idea which I tried to make a quick mock up of:
   screen shot 2018-01-07 at 7 54 52 pm
   - I like it with nothing in the middle, but would like to see versions with an 'S', or the redux logo (I think this one will look too busy)
   - lines should be same width and length, and evenly spaced
   - prefer the weight of the inner layer to the outer layer
   - play with rotation of layers to feel less symmetrical

If someone with more skills in making images want to tidy this up I'd love to see it.  I do very much like @matthewhartman's direction, so don't think my mock up is the final word, just an idea I had.

mpeyper:
I'm going to close this as it's been pretty stale for a while.

Happy to reopen if anyone wants to continue with it, or feel free to open a new PR (or just reply to #60 if you want to discuss ideas).

PR #62, opened 2018-01-11
Fix issue with multiple applyMiddleware enhancers not being applied  to subspaces

lkyles1991:
Not quite sure if the test is in the right part or if the test should be adapted to use the middware/reducers that currently exist in the test.

### DIFF
--- a/packages/redux-subspace/test/integration-spec.js
+++ b/packages/redux-subspace/test/integration-spec.js
@@ -8,7 +8,7 @@
 
 import path from 'path'
 import fs from 'fs'
-import { createStore, combineReducers } from 'redux'
+import { createStore, combineReducers, compose } from 'redux'
 import { 
     subspace,
     applyMiddleware,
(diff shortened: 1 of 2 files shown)


packages/redux-subspace/test/integration-spec.js
```diff
@@ -831,6 +831,50 @@ describe('integration tests', () => {
...
+        
+            const middleware = (newKey, value) => () => (next) => (action) => {
+                if (action.type === TEST_ACTION) {
+                    return next({ ...action, [newKey]: value })
+                } 
+                return next(action)
+            }
+
+            const parentStore = createStore(parentReducer, compose(
+            applyMiddleware(middleware('value2', 'middleware value 1')),
```
mpeyper: huge nitpick, but could you indent these lines please

lkyles1991:
Added the indentation to the test. No idea how I never spotted that

PR #63, opened 2018-01-12
Pass saga args through to subspaced sagas

mpeyper:
redux-saga supports [passing arguments to sagas when running them](https://redux-saga.js.org/docs/api/#middlewarerunsaga-args).  Although `createSagaMiddleware` does pass the args on when to the underlying `run` call, the `subspaced` and `provideStore` higher-order sagas would not pass them through to the sagas they were wrapping.

### DIFF
--- a/packages/redux-subspace-saga/src/sagas/provideStore.js
+++ b/packages/redux-subspace-saga/src/sagas/provideStore.js
@@ -8,12 +8,11 @@
 
 import { setContext } from 'redux-saga/effects'
 
-const provideStore = (store, options) => (saga) => {
-    return function* () {
-        yield setContext({ store })
-        yield setContext({ sagaMiddlewareOptions: options })
-        yield* saga()
-    }
+const provideStore = (store, options) => saga => {
+  return function*(...args) {
+    yield setContext({ store, sagaMiddlewareOptions: options })
+    yield* saga(...args)
+  }
 }
 
 export default provideStore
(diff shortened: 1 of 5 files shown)


packages/redux-subspace-saga/test/integration-spec.js
```diff
@@ -43,113 +43,9 @@ describe('integration tests', () => {
         }
     }
 
-    it('should transfer context to subspaced saga', () => {
```
mpeyper: I only moved the context tests down the file.

packages/redux-subspace-saga/test/integration-spec.js
```diff
@@ -366,4 +262,200 @@ describe('integration tests', () => {
...
+                child2: 'initial value'
+            },
+            parent2: {
+                child1: 'initial value',
+                child2: 'context value'
+            }
+        })
+    })
+
+    it('should pass args through to subspaced saga', () => {
```
mpeyper: This is where the new tests for passing args start

packages/redux-subspace-saga/src/sagas/provideStore.js
```diff
@@ -8,12 +8,11 @@
 
 import { setContext } from 'redux-saga/effects'
 
-const provideStore = (store, options) => (saga) => {
-    return function* () {
-        yield setContext({ store })
-        yield setContext({ sagaMiddlewareOptions: options })
```
mpeyper: There was no reason to make seperate calls to `setContext` here.

jpeyper:
Looks good to me :)

PR #66, opened 2018-01-18
Fix deprecation warning by injecting store as dependency

### LINKED ISSUE #65: [redux-subspace-observable] DEPRECATION message is trowed by redux-observable
For such simple example:

```
import 'rxjs/add/operator/map';
import { compose, createStore } from 'redux';
import { createEpicMiddleware, subspaced } from 'redux-subspace-observable';
import { applyMiddleware,  } from 'redux-subspace';

let myEpic = (actions$) => actions$.ofType('A').map(a => {return { type: 'B'};});
let mySubspacedEpic = subspaced(state => state['nested'], 'nested')(myEpic);
let reducer = (state = {}) => {return state; };

let store = createStore(
    reducer,
    compose(applyMiddleware(createEpicMiddleware(mySubspacedEpic)))
);

store.dispatch({type: 'nested/A'});
```
the DEPRECATION message is trowed by  'redux-observable'
```
redux-observable | DEPRECATION: calling store.dispatch() directly in your Epics is deprecated and will be removed. Instead, emit actions through the Observable your Epic returns.
https://goo.gl/WWNYSP
```

```
    "redux": "^3.7.2",
    "redux-observable": "^0.17.0",
    "redux-subspace": "^2.1.0",
    "redux-subspace-observable": "^2.1.0",
    "rxjs": "^5.5.6",
```

Best regards.

P.S.
Many thx for your effort on great lib :)

evertbouw:
closes #65

### DIFF
--- a/packages/redux-subspace-observable/src/index.js
+++ b/packages/redux-subspace-observable/src/index.js
@@ -11,4 +11,7 @@ import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observab
 
 export { default as subspaced } from './observable/subspaced'
 
-export const createEpicMiddleware = (rootEpic, options) => applyToRoot(baseCreateEpicMiddleware(rootEpic, options))
+export const createEpicMiddleware = (rootEpic, options = {}) =>
+    applyToRoot(store =>
+        baseCreateEpicMiddleware(rootEpic, { ...options, dependencies: { ...options.dependencies, store } })(store)
+    )
--- a/packages/redux-subspace-observable/src/index.js
+++ b/packages/redux-subspace-observable/src/index.js
@@ -6,12 +6,5 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { applyToRoot } from 'redux-subspace'
-import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observable'
-
 export { default as subspaced } from './observable/subspaced'
-
-export const createEpicMiddleware = (rootEpic, options = {}) =>
-    applyToRoot(store =>
-        baseCreateEpicMiddleware(rootEpic, { ...options, dependencies: { ...options.dependencies, store } })(store)
-    )
+export { createEpicMiddleware } from './observable/subspaceEpicMiddleware'
--- a/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
+++ b/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
@@ -1,7 +1,15 @@
 import { applyToRoot } from 'redux-subspace'
 import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observable'
+import { SUBSPACE_STORE_KEY } from './subspaceStoreKey';
 
 export const createEpicMiddleware = (rootEpic, options = {}) =>
-    applyToRoot(store =>
-        baseCreateEpicMiddleware(rootEpic, { ...options, dependencies: { ...options.dependencies, store } })(store)
-    )
+    applyToRoot(store => {
+        const optionsWithStore = {
+            ...options,
+            dependencies: {
+                ...options.dependencies,
+                [SUBSPACE_STORE_KEY]: store
+            }
+        }
+        return baseCreateEpicMiddleware(rootEpic, optionsWithStore)(store)
+    })
--- a/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
+++ b/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
@@ -2,8 +2,13 @@ import { applyToRoot } from 'redux-subspace'
 import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observable'
 import { SUBSPACE_STORE_KEY } from './subspaceStoreKey';
 
-export const createEpicMiddleware = (rootEpic, options = {}) =>
-    applyToRoot(store => {
+export const createEpicMiddleware = (rootEpic, options = {}) => {
+    if (process.env.NODE_ENV !== 'production') {
+        const dependenciesIsObject = options.dependencies == undefined || typeof options.dependencies === 'object'
+        console.assert(dependenciesIsObject, 'dependencies must be an object')
+    }
+
+    return applyToRoot(store => {
         const optionsWithStore = {
             ...options,
             dependencies: {
@@ -13,3 +18,4 @@ export const createEpicMiddleware = (rootEpic, options = {}) =>
         }
         return baseCreateEpicMiddleware(rootEpic, optionsWithStore)(store)
     })
+}
--- a/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
+++ b/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
@@ -3,9 +3,8 @@ import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observab
 import { SUBSPACE_STORE_KEY } from './subspaceStoreKey';
 
 export const createEpicMiddleware = (rootEpic, options = {}) => {
-    if (process.env.NODE_ENV !== 'production') {
-        const dependenciesIsObject = options.dependencies == undefined || typeof options.dependencies === 'object'
-        console.assert(dependenciesIsObject, 'dependencies must be an object')
+    if (options.dependencies != null && (Array.isArray(options.dependencies) || typeof options.dependencies !== 'object')) {
+        throw new TypeError('dependencies must be an object')
     }
 
     return applyToRoot(store => {
--- a/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
+++ b/packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
@@ -1,0 +1,7 @@
+import { applyToRoot } from 'redux-subspace'
+import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observable'
+
+export const createEpicMiddleware = (rootEpic, options = {}) =>
+    applyToRoot(store =>
+        baseCreateEpicMiddleware(rootEpic, { ...options, dependencies: { ...options.dependencies, store } })(store)
+    )
--- a/packages/redux-subspace-observable/src/observable/subspaced.js
+++ b/packages/redux-subspace-observable/src/observable/subspaced.js
@@ -16,13 +16,15 @@ const identity = (x) => x
 const subspaced = (mapState, namespace) => {
     const subspaceDecorator = subspace(mapState, namespace)
     return epic => (action$, store, dependencies) => {
-        const subspacedStore = subspaceDecorator(store)
+        const subspacedStore = subspaceDecorator(dependencies.store)
 
         const filteredAction$ = action$
             ::map((action) => subspacedStore.processAction(action, identity))
             ::filter(identity)
 
-        epic(filteredAction$, subspacedStore, dependencies).subscribe(subspacedStore.dispatch)
+        epic(filteredAction$, subspacedStore, { ...dependencies, store: subspacedStore })
+            .subscribe(subspacedStore.dispatch)
+
         return empty()
     }
 }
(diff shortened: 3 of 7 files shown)


jayphelps:
LGTM (not a maintainer of this lib, just asked to review)

packages/redux-subspace-observable/src/index.js
```diff
@@ -11,4 +11,7 @@ import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observab
 
 export { default as subspaced } from './observable/subspaced'
 
-export const createEpicMiddleware = (rootEpic, options) => applyToRoot(baseCreateEpicMiddleware(rootEpic, options))
+export const createEpicMiddleware = (rootEpic, options = {}) =>
```
mpeyper: @evertbouw @jayphelps, is it safe to assume dependencies is an object, or should we need to cater for cases where something else (array, primitive, function, etc.) are passed?

packages/redux-subspace-observable/test/observable/subspaced-spec.js
```diff
@@ -134,7 +144,7 @@ describe('subspaced Tests', () => {
         ])
     })
     
-    it('should pass on dependeies to epic', () => {
+    it('should pass on dependencies to epic', () => {
```
mpeyper: :)

packages/redux-subspace-observable/src/observable/subspaced.js
```diff
@@ -16,15 +16,17 @@ const identity = (x) => x
 const subspaced = (mapState, namespace) => {
     const subspaceDecorator = subspace(mapState, namespace)
     return epic => (action$, store, dependencies) => {
-        const subspacedStore = subspaceDecorator(store)
+        const subspacedStore = subspaceDecorator(dependencies.store)
```
mpeyper: might want some validation on whether `dependencies` is an object and it has a `store` field (i.e. they have now used our `createEpicMiddleware` wrapper) for development

packages/redux-subspace-observable/src/index.js
jayphelps: hehe I thought about the same thing but didn't mention it because, while nothing in redux-observable enforces it being an object, not doing so would be weird and discouraged.

jayphelps: not saying you can't be more defensive and assert that it is 😄

packages/redux-subspace-observable/src/observable/subspaced.js
```diff
@@ -16,15 +16,17 @@ const identity = (x) => x
...
 
         const filteredAction$ = action$
             ::map((action) => subspacedStore.processAction(action, identity))
             ::filter(identity)
 
-        epic(filteredAction$, subspacedStore, dependencies).subscribe(subspacedStore.dispatch)
+        epic(filteredAction$, subspacedStore, { ...dependencies, store: subspacedStore })
```
mpeyper: I wonder if we should be a bit more careful with the `store` param we pass on to the epic.  With this change the `subspaced` epic would still be able to dispatch into the store directly, bypassing the warning that was introduced.  Assuming this is still seen as an anti-pattern, I don't think we should swallow the warning.  This is more important for v1.0.0, where the dispatch function is expected to be missing, but a `subspaced` epic will still be able to dispatch happily.

Thoughts?

mpeyper:
Thanks @evertbouw for this.  I've made a few comments on some improvements, but I think this is fine to merge now to remove the deprecation warning for our users.

I thinkhave made should be done sooner rather than later (not necessarily by you if you don't want).  Perhaps I'll raise a few `good first issue` issues and give some new folks a chance to make a contribution.

I'll wait for one more tick from a team member, just to make sure I haven't missed anything obvious.

packages/redux-subspace-observable/src/observable/subspaced.js
jayphelps: Super great points!

As far as the second argument store `(action$, store, deps)` and not preventing the deprecation warning if they themselves call `store.dispatch`..I'm not sure how this will be possible. You may have to just duplicate similar deprecation logic lol.

Separately but related: I wouldn't pass on the subspacedStore in dependencies to users epics. The fact that you internally pass the store along dependencies should just be an implementation detail that doesn't impact users in any way. That would also mean that if users themselves pass a dependency with the key `store` it should still get passed along no matter what it is. I've helped some people do just that to temporarily future-proof their epics from the store's removal--its still discouraged, but at least then people are knowingly opting into the anti-pattern instead of us giving them the gun.

I think the solution to that is to store the users provided dependencies in a specific property instead of spreading it.

```js
baseCreateEpicMiddleware(rootEpic, { ...options, dependencies: { store, whatever: options.dependencies } })(store)
```

Then it can be extracted as-is and passed along to the users epics without knowing what type it is and not leaking the store in it.

jayphelps: Whatever solution it won't have to live long. The second argument is changing to state$ as a stream in the coming weeks in v1.0.0-alpha

mpeyper: @jayphelps 

> You may have to just duplicate similar deprecation logic lol.

Yeah, I'm thinking the same, but checking if `dispatch` is there first (i.e. if dispatch is on the store param, add a deprecation wrapper).  This was we will be compatible with current versions and future versions (we would probably remove it all together one day).

> I wouldn't pass on the subspacedStore in dependencies to users epics
> ...
> I think the solution to that is to store the users provided dependencies in a specific property instead of spreading it.

One of the features of redux subspace is being able to nest subspaces arbitrarily deep.  Because of this, we the nested subspace needs access to it's parent's subspace and use it the same way the parent subspace used the root store.

We also don't want to disrupt the `dependencies`that are passed too much as the rootEpic itself might have non-subspaced epics combined in.

For example:

```
const nestedEpic = (action$, store, dependencies) => { // uses dependencies.api }
const subspacedNestedEpic = subspaced((state) => state.nestedState, 'nested')(nestedEpic)

const standardMiddleEpic = (action$, store, dependencies) => { // uses dependencies.api }
const combinedMiddleEpic = combineEpics(standardMiddleEpic, subspacedNestedEpic)
const subspacedMiddleEpic = subspaced((state) => state.middleState, 'middle')(combinedMiddleEpic)

const standardRootEpic = (action$, store, dependencies) => { // uses dependencies.api }
const rootEpic = combineEpics(standardRootEpic, subspacedMiddleEpic)

const epicMiddleware = createEpicMiddleware(rootEpic, { dependencies: { api: { // create api } } })
const store = createStore(rootReducer, applyMiddleware(epicMiddleware))
```

Does that make sense?

> redux-observable won't pass undefined as the dependencies if they didn't pass the dependencies key in the options

I don't think this will be an issue for us as as far as redux-observable is concerned, there was always `dependencies` in the `options`.

> The second argument is changing to state$ as a stream in the coming weeks in v1.0.0-alpha

I saw this when looking for solutions to for this bug.  I'm not sure what this means for us.  I think we will need to map the `state$` stream and run the `mapState` function over it before passing it to the wrapped epic.  Does that sound feasible?

evertbouw:
I should focus on my day job tomorrow but I can do it this weekend. Leaving it open for other folks would also work for me :shipit:

packages/redux-subspace-observable/src/observable/subspaced.js
jayphelps: > We also don't want to disrupt the dependencies that are passed too much as the rootEpic itself might have non-subspaced epics combined in.
> Does that make sense?

This one lost me actually. 😄  My suggestion was to not touch their dependencies which 

> I think we will need to map the state$ stream and run the mapState function over it before passing it to the wrapped epic. Does that sound feasible?

Not 100% sure what you mean but based on the code I've seen there shouldn't be a problem coming up with a solution. You'll have access to a stream of actions, a stream of state, and the latest state imperatively `state$.value` so I believe everything you would possibly need.

mpeyper: > This one lost me actually. 😄 My suggestion was to not touch their dependencies which sounds the same as "don't want to disrupt the dependencies" to me lol

Yep, I can see why you're confused...  What I meant was we don't want to change the path the the existing dependencies as there might be a epic that has been combined that is not in a `subspaced` wrapper.

So in my (contrived) example, `standardRootEpic` expects to be able to access `api` directly on the dependencies object.  If we moved them into a `whatever` node as you suggested, this epic would need to know that and change to use `dependencies.whatever.api`.

jayphelps: @mpeyper oh gotcha! hmm. I would place the store on a more private prop name then e.g. `__storeForSubspaceDontUseOrYouWillBeFired` to prevent it clashing with whatever they might provide for `store`. Chances are it would be the same store, but seems easy enough to avoid anyway.

jayphelps: Prolly also noneumerable

```js
Object.defineProperty(dependencies, '__storeForSubspaceDontUseOrYouWillBeFired', {
  enumerable: false, // most important one to be a good citizen
  configurable: false,
  writable: false,
  value: store
});
```

jayphelps: Anyways, sorry to jump in and start pushing my pedanticism 🕺

mpeyper: So many things I did not know were even possible!

I think the `...DontUseOrYouWillBeFired` is a bit overkill.  Maybe something like `@@redux-subspace/store` so it doesn't even look like a key that could be used (I tested in jsfiddle and it seems to be allowed with `defineProperty`.

I think it will still need to writable so it can be replaced for nested subspaces, but other than that I like the idea.  Maybe we should do this in more places redux-subspace? 🤔

jayphelps: @mpeyper lolol the prop name is totally your call, the name was just a joke based on the React one haha 😄 

https://github.com/facebook/react/blob/b2ca3349c27b57b1e9462944cbe4aaaf76783d2b
/src/React.js#L67

> React.__SECRET_DOM_DO_NOT_USE_OR_YOU_WILL_BE_FIRED = ReactDOM;

evertbouw:
I'm also not sure about that. If the user provides something that is not an object it wouldn't be a problem for redux-subspace, but weird stuff can happen in their epics. Primitives would be ignored but arrays will be spread to an object.

packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
```diff
@@ -0,0 +1,21 @@
+import { applyToRoot } from 'redux-subspace'
+import { createEpicMiddleware as baseCreateEpicMiddleware } from 'redux-observable'
+import { SUBSPACE_STORE_KEY } from './subspaceStoreKey';
+
+export const createEpicMiddleware = (rootEpic, options = {}) => {
+    if (process.env.NODE_ENV !== 'production') {
+        const dependenciesIsObject = options.dependencies == undefined || typeof options.dependencies === 'object'
+        console.assert(dependenciesIsObject, 'dependencies must be an object')
```
jayphelps: How about an exception, or a `console.warn`? `console.assert` has inconsistent behavior across browsers and node--in some it throws, in others it just logs an error but doesn't throw. Node is used when people server-side render. Personally I would throw, like you did in https://github.com/ioof-holdings/redux-subspace/pull/66/files#diff-d6110079ec3df7516f551a2c9957fad6R20

mpeyper: The things you're teaching me @jayphelps.... we've used `console.assert` as the default way to raise errors within redux-subspace.  The intention was to throw, so if `assert` is not consistent at doing that, we should change them all.

I did wonder why other libraries seemed not to use it as it seemed perfect for what I wanted.

mpeyper:
> it wouldn't be a problem for redux-subspace, but weird stuff can happen in their epics

I think we should be good citizens and prevent that weirdness.

packages/redux-subspace-observable/src/observable/subspaceEpicMiddleware.js
jpeyper: Lets fix them all in another PR.

mpeyper: Agreed. Well beyond the scope of this change.

mpeyper:
Sorry it's taken me so long to get back to this.  You've gone above and beyond on this one @evertbouw.  Thanks again!

PR #74, opened 2018-01-28
Change all console.asserts to throw TypeErrors

### LINKED ISSUE #72: [react-native] console.assert is not a function
It looks like when react-native app is running without remote debugger,  app crash due to lack of assert function in the console object. I would suggest adding polyfill, or just check if assert is available before running it.

jpeyper:
Resolves #72


mpeyper:
LGTM 💯

PR #75, opened 2018-01-30
middleware api now includes all subspace values so root enhancer can be applied in middleware

jpeyper:
subspaceOptions (which has the root enhancer) was not being passed through to the middleware API. Any subspaces created in middleware were not correctly applying the middleware chain.

I felt it's safer to to exclude the parts of the subspacedStore we don't want, to avoid issues like this in the future (we have to be careful to remember to update the typescript definitions).


mpeyper:
This has been causing some absolute doozey bugs... Glad we found it!

Issue #76, opened 2018-01-31
Subspaced epic should re-throw an error 


psamusev:
I would like to add error handling for all of my epics for this I use
```javascript
epic(action, store, deps).pipe(
     catchError((error, stream) => {
         / custom error handling
         return stream;   // return stream to make epic alive after any error
})
```

when I use it outside of subspace all of my errors appear in this ** catchError** callback.
But when epic is executed from subspace epic error is swallowing and **catchError** is not executing. By this epic is dead but I need keep it alive. 

epic which is returning by subspaced should re-throw an error if it happened in original epic

mpeyper:
Given [how the `subspaced` wrapper is implemented](https://github.com/ioof-holdings/redux-subspace/blob/master/packages/redux-subspace-observable/src/observable/subspaced.js#L39), it's not surprising that an error raised in the original epic cannot be caught from the outside.

I'm not very experienced with using rx myself so can you please elaborate on the use case here?  I'm trying to understand what you might be doing in your custom error handling and what "return stream to make epic alive after any error" means.

My concern is that if the `subspaced` wrapper emits the error and you handle it, then the epic running inside the wrapper is too isolated to get any benefit.

If you have any ideas on how `subspaced` can be modified to better meet your requirements, I'm more than happy to discuss that here or in a PR.

psamusev:
Ok, these are several examples. How error influence on Observer:

1) If inside it you coutch an error Observer would terminate - it means that sequence of this observable is no more handling.

For instance we have an epic
```javascript
const epic = (action, store) => 
action.ofType(SOME_TYPE)
     .mergeMap((action) => {
              if (someFlag) {
                        return Observable.of(value)
              }
              throw new Error('Wrong flag value');
     })
```
if we catch an error whole epic will be terminated. It means that during one more event dispatch of this event - SOME_TYPE - nothing will be happened

2) If inside it you coutch an error and even handled it with *catch* Observer would terminate as well.

For instance we have an epic
```javascript
const epic = (action, store) => 
action.ofType(SOME_TYPE)
     .mergeMap((action) => {
              if (someFlag) {
                        return Observable.of(value)
              }
              throw new Error('Wrong flag value');
     })
     .catch((error) => {  // do some error handling })
```
if we catch an error whole epic will be terminated as well even with handling.

3) If inside it you coutch an error and handled it with *catch* and one important thing return the same stream epic will continue leave.

For instance we have an epic
```javascript
[...]

PR #77, opened 2018-02-02
fixing #76

majo44:
fixing #76


mpeyper:
Thanks for this and all your help with #76 💯

PR #80, opened 2018-02-17
Add a Code of Conduct

mpeyper:
I've wanted to add a code of conduct for the repo for a while now.

This is the [Contributor Covenant](https://www.contributor-covenant.org/) which is a pretty standard code of conduct for OSS projects.

My only concern is that it requires an email address for reporting breaches.  I've used my IOOF email for this (I'm happy to receive them), but I wonder if there should be a more official email used for this?


mpeyper:
Well, no one is jumping up and down about this, so I'm going to go ahead and merge it.

PR #85, opened 2018-04-15
Annotate pure calls with babel plugin to slightly improve tree-shakea…

### LINKED ISSUE #84: Add `#__PURE__` comments to packages to improve tree shaking
### Is it a bug, feature request or question?

In [PR to build flat bundles](https://github.com/ioof-holdings/redux-subspace/pull/78#issue-168415107) @Andarist mentioned a [babel plugin](https://github.com/Andarist/babel-plugin-annotate-pure-calls) he wrote to improve the tree shakability (yes, I just made that word up) of the outputted code.

I would like to include it in our babel configurations for each of the packages.

### Which package(s) does this involve?

All packages.

Andarist:
…bility

| Q                       | A 
| Fixed Issues?           | Fixes #84,
| Documentation only PR   | 
| Patch: Bug Fix?         | 
| Minor: New Feature?     | 👍
| Major: Breaking Change? | 

Truth to be told only 2 packages **at the moment** benefit from it.

`redux-subspace-loop` - `namespaced` export is created by a call and it might get dropped if the package stays unused

`react-redux-subspace` - `SubspaceProvider` gets correctly annotated, but it cannot be removed because it uses static properties, so the output looks like this:
```js
var SubspaceProvider = /*#__PURE__*/function (_React$PureComponent) {
 // ...
}(React__default.PureComponent);

SubspaceProvider.propTypes = {
    children: PropTypes.element.isRequired,
    mapState: /*#__PURE__*/PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
    namespace: PropTypes.string,
    subspaceDecorator: PropTypes.func
};

SubspaceProvider.contextTypes = {
    store: PropTypes.object.isRequired
};

SubspaceProvider.childContextTypes = {
    store: PropTypes.object
};
```

Unfortunately such static assignments prevent dead code removal, because algorithm thinks that `SubspaceProvider` is used.

There are 2 ways of "fixing" this problem:
- custom babel plugin to pull those static assignments into the transpiled class IIFE, I might create it some day, but in general would like to fix it upstream - I have working PR for this in babel@7, but it's not merged in et
- create this class from the IIFE, like this:
```js
const SubspaceProvider = (() => {
  class SubspaceProvider extends React.PureComponent {
    // ...
  }
  return SubspaceProvider
})()
export default SubspaceProvider
```
this will let the plugin annotate outer IIFE (inner being a transpiled class) which would contain whole class definition and thus making it dead code droppable.

But ofc this is somewhat ugly and also gives rather rly small benefits - in real world scenarios for this library probably even none.

### DIFF
--- a/packages/react-redux-subspace/.babelrc.js
+++ b/packages/react-redux-subspace/.babelrc.js
@@ -9,6 +9,7 @@ module.exports = {
     'stage-3',
   ],
   plugins: [
+    'annotate-pure-calls',
     cjs && 'transform-es2015-modules-commonjs',
   ].filter(Boolean),
 }
--- a/packages/react-redux-subspace/package-lock.json
+++ b/packages/react-redux-subspace/package-lock.json
@@ -1,11 +1,14 @@
 {
-	"requires": true,
+	"name": "react-redux-subspace",
+	"version": "2.4.0",
 	"lockfileVersion": 1,
+	"requires": true,
 	"dependencies": {
 		"@babel/code-frame": {
 			"version": "7.0.0-beta.40",
 			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.40.tgz",
 			"integrity": "sha512-eVXQSbu/RimU6OKcK2/gDJVTFcxXJI4sHbIqw2mhwMZeQ2as/8AhS9DGkEDoHMBBNJZ5B0US63lF56x+KDcxiA==",
+			"dev": true,
 			"requires": {
 				"@babel/highlight": "7.0.0-beta.40"
 			}
@@ -14,6 +17,7 @@
 			"version": "7.0.0-beta.40",
 			"resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.0.0-beta.40.tgz",
 			"integrity": "sha512-c91BQcXyTq/5aFV4afgOionxZS1dxWt8OghEx5Q52SKssdGRFSiMKnk9tGkev1pYULPJBqjSDZU2Pcuc58ffZw==",
+			"dev": true,
 			"requires": {
 				"@babel/types": "7.0.0-beta.40",
 				"jsesc": "2.5.1",
@@ -25,7 +29,8 @@
 				"jsesc": {
 					"version": "2.5.1",
 					"resolved": "https://registry.npmjs.org/jsesc/-/jsesc-2.5.1.tgz",
-					"integrity": "sha1-5CGiqOINawgZ3yiQj3glJrlt0f4="
+					"integrity": "sha1-5CGiqOINawgZ3yiQj3glJrlt0f4=",
+					"dev": true
 				}
 			}
 		},
@@ -33,6 +38,7 @@
 			"version": "7.0.0-beta.40",
 			"resolved": "https://registry.npmjs.org/@babel/helper-function-name/-/helper-function-name-7.0.0-beta.40.tgz",
 			"integrity": "sha512-cK9BVLtOfisSISTTHXKGvBc2OBh65tjEk4PgXhsSnnH0i8RP2v+5RCxoSlh2y/i+l2fxQqKqv++Qo5RMiwmRCA==",
+			"dev": true,
 			"requires": {
 				"@babel/helper-get-function-arity": "7.0.0-beta.40",
 				"@babel/template": "7.0.0-beta.40",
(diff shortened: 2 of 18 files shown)


mpeyper:
As long the core package (`redux-subspace`) gets some benefit, I'm happy.  That's the one with the most stuff that isn't going to be used by everyone.

I'll try and find some time tonight to take a closer look, but on the surface everything looks fine here.

Andarist:
well, `redux-subspace` does not benefit from this at the moment, because nothing in top level scope is created with a function call - so regular tree-shaking should be able to remove unused stuff anyway

this can ofc change in the future, so it might be good to leave the plugin just in case

mpeyper:
> `redux-subspace` does not benefit from this at the moment, because nothing in top level scope is created with a function call

hmmm, ok.  I think I understand a bit better what this plugin doing now.  When a variable is assigned from a function call just by importing a file, this tells the tree shaker that is was a pure call and can be removed.  Without it, the variable can be shaken away, but the call, and consequently function itself, are left behind as it may have had a side-effect worth keeping.  Am I close?

> `SubspaceProvider` gets correctly annotated, but it cannot be removed because it uses static properties

Changing `SubspaceProvider to

```js
class SubspaceProvider extends React.PureComponent {
  static get propTypes() {
    return {
      children: PropTypes.element.isRequired,
      mapState: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
      namespace: PropTypes.string,
      subspaceDecorator: PropTypes.func
    }
  }

  static get contextTypes() {
    return {
      store: PropTypes.object.isRequired
    }
  }

  static get childContextTypes() {
    return {
      store: PropTypes.object
    }
  }

  getChildContext() {
    const makeSubspaceDecorator = props => props.subspaceDecorator || subspace(props.mapState, props.namespace)

    return {
      store: makeSubspaceDecorator(this.props)(this.context.store)
    }
  }

  render() {
    return Children.only(this.props.children)
  }
}
```

outputs

```js
var SubspaceProvider = /*#__PURE__*/function (_React$PureComponent) {
  inherits(SubspaceProvider, _React$PureComponent);

  function SubspaceProvider() {
    classCallCheck(this, SubspaceProvider);
    return possibleConstructorReturn(this, _React$PureComponent.apply(this, arguments));
  }

  SubspaceProvider.prototype.getChildContext = function getChildContext() {
    var makeSubspaceDecorator = function makeSubspaceDecorator(props) {
      return props.subspaceDecorator || subspace(props.mapState, props.namespace);
    };

    return {
      store: makeSubspaceDecorator(this.props)(this.context.store)
    };
  };

  SubspaceProvider.prototype.render = function render() {
    return Children.only(this.props.children);
  };

  createClass(SubspaceProvider, null, [{
    key: 'propTypes',
    get: function get$$1() {
      return {
        children: PropTypes.element.isRequired,
        mapState: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
        namespace: PropTypes.string,
        subspaceDecorator: PropTypes.func
      };
    }
  }, {
    key: 'contextTypes',
    get: function get$$1() {
      return {
        store: PropTypes.object.isRequired
      };
    }
  }, {
    key: 'childContextTypes',
    get: function get$$1() {
      return {
        store: PropTypes.object
      };
    }
  }]);
  return SubspaceProvider;
}(React.PureComponent);
```

which I think means it can be shaken away, but I'm not sure if that's necessarily a recommended (or valid?) way to define a React component. 

> this can ofc change in the future, so it might be good to leave the plugin just in case

I like the idea of helping future me (and others) out and leaving it in for all packages, even if they don;t get any immediate benefit from it right now.

Andarist:
> hmmm, ok. I think I understand a bit better what this plugin doing now. When a variable is assigned from a function call just by importing a file, this tells the tree shaker that is was a pure call and can be removed. Without it, the variable can be shaken away, but the call, and consequently the function itself, are left behind as it may have had a side-effect worth keeping. Am I close?

You couldn't be closer :)

> which I think means it can be shaken away, but I'm not sure if that's necessarily a recommended (or valid?) way to define a React component?

Neat trick! didn't think of that, although I'm not sure if using getters for those is a good thing.

> I like the idea of helping future me (and others) out and leaving it in for all packages, even if they don't get any immediate benefit from it right now.

👍

mpeyper:
> I'm not sure if using getters for those is a good thing.

No, me neither.  I don't think I'll worry with it.

The reality is, whether you use `subspaced` or `SubspaceProvider` from that package, you will need that code, so it's unlikely it would ever have been removed if the package was used.

I'm happy with this and will merge it now.  Thanks again :)

PR #86, opened 2018-04-19
Use pipe operators for rxjs

evertbouw:
Hey 👋 

I was playing around with this lib to see what it would take to get it compatible with redux-observable v1. I'll report back when I've go something, but in the mean time I've replaced the patch operators with the rxjs 6 style pipe operators.

This also removes the need to use the function bind plugin for babel.

### DIFF
--- a/packages/redux-subspace-observable/package.json
+++ b/packages/redux-subspace-observable/package.json
@@ -34,11 +34,11 @@
     "@types/es6-shim": "^0.31.34",
     "@types/node": "^8.0.47",
     "redux-subspace": "^2.4.0",
-    "rxjs": "^5.5.2"
+    "rxjs": "^5.5.10"
   },
   "peerDependencies": {
-    "redux": "^3.0.0",
-    "redux-observable": "^0.16.0"
+    "redux": ">3 <4",
+    "redux-observable": ">0.16 <1"
   },
   "devDependencies": {
     "babel-cli": "^6.18.0",
@@ -46,7 +46,6 @@
     "babel-eslint": "^8.0.1",
     "babel-plugin-annotate-pure-calls": "^0.2.2",
     "babel-plugin-external-helpers": "^6.22.0",
-    "babel-plugin-transform-function-bind": "^6.22.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.5.0",
     "babel-preset-stage-3": "^6.24.1",
@@ -59,7 +58,7 @@
     "nyc": "^11.3.0",
     "redux": "^3.6.0",
     "redux-mock-store": "^1.3.0",
-    "redux-observable": "^0.17.0",
+    "redux-observable": "^0.18.0",
     "rollup": "^0.55.5",
     "rollup-plugin-babel": "^3.0.3",
     "sinon": "^4.0.2",
--- a/packages/redux-subspace-observable/.babelrc.js
+++ b/packages/redux-subspace-observable/.babelrc.js
@@ -10,7 +10,6 @@ module.exports = {
   ],
   plugins: [
     'annotate-pure-calls',
-    cjs && 'transform-es2015-modules-commonjs',
-    'transform-function-bind'
+    cjs && 'transform-es2015-modules-commonjs'
   ].filter(Boolean),
 }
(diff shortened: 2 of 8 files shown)


mpeyper:
Hello again 😄 

I had a go at using the `pipe` function for this recently but couldn't get it working (I swear it was identical to yours now 😕).

This looks like a minor version change to me, can you think of any reason why it would be a breaking change for consumers?

While I've got you, I stumbled across [this comment](https://gitter.im/redux-observable/redux-observable?at=5a6103bd9cdc721e4f9b8cbe) by @jayphelps.  From my (limited) understanding of `rxjs` and `Observables`, it sounds like we might not be as [replaceable with real redux stores](https://github.com/reactjs/redux/blob/master/src/createStore.js#L268) as we thought, and we could more easily support redux-observable if subspaces has a have `Symbol.observable`.  Does any of that make sense?

packages/redux-subspace-observable/package.json
```diff
@@ -34,19 +34,18 @@
     "@types/es6-shim": "^0.31.34",
     "@types/node": "^8.0.47",
     "redux-subspace": "^2.4.0",
-    "rxjs": "^5.5.2"
+    "rxjs": "^5.5.10"
   },
   "peerDependencies": {
     "redux": "^3.0.0",
-    "redux-observable": "^0.16.0"
+    "redux-observable": "^0.18.0"
```
jayphelps: Traditionally this would technically make it a breaking change.

jayphelps:
@mpeyper I don't believe the issue described in that gitter thread is problem anymore due to https://github.com/ioof-holdings/redux-subspace/pull/66. On an unrelated note it wouldn't hurt to also support Symbol.observable though.

I believe the current version of subspace and the current version of redux-observable are both compatible, but the issue is that upcoming redux-observable v1 (and the current 1.0.0-alpha.2) are not because it requires rxjs v6. This PR appears to get us closer to supporting it by moving to the pipeable operators, thought still on rxjs v5.

evertbouw:
I think there will be some more breaking changes coming up with Redux 4, if you want to make a release with this in it probably won't last too long.

This change can have a positive or negative effect on bundle size, depending on what operators the consuming project uses

packages/redux-subspace-observable/package.json
mpeyper: Is that because alpha versions are can be breaking with minor bumps?  If it was, for example `^1.16.0` -> `1.18.0`, would that also traditionally be considered breaking? (I have never thought of minor bumps in dependencies as breaking changes).

jayphelps: My understanding is that if someone is following semver, pre 1.0 bumps of any kind should be considered unstable, but that maintainers will try to use minor bumps to try and gate potentially breaking changes. So 0.16.0 -> 0.18.0 could be breaking. But 1.16.0 -> 1.18.0 should not be. In this specific case of redux-observable there [weren't any known breaking changes between 0.16.0 and 0.18.0](https://github.com/redux-observable/redux-observable/blob/master/CHANGELOG.md#0180-2018-02-07) other than type definition changes, which IMO are almost always "breaking", however I would still consider this peer dep bump breaking as if a consumer has locked their redux-observable version to 0.16.0, they'll now get a peer deps error on install, even though they could probably fairly easily upgrade.

Unless there is a big reason against, I would personally advocate for subspace skipping 0.18.0 and just working towards v1.0.0 and rxjs v6.

evertbouw: I agree with that. This PR is mostly just prep for that

evertbouw:
Just noticed the lockfile was generated with our internal npm registry 🤦‍♂️ took the opportunity to change the peerDep range

mpeyper:
Sorry, I've been absent the last few days.  I've caught up on the comments, but I'm a bit confused.  I'm conceptually happy with the changes, so should I be merging this now and pushing out a release, or am I leaving this open for now and waiting for compatibility with redux-observable v1.0.0 and rxjs v6?

evertbouw:
That's up to you I suppose. Making a release with these changes could help people who are migrating to pipe operators in their own code.

mpeyper:
Sorry again for my absence (my wife and I are about a week away from having a baby).

I'll merge this and work on getting it released soon.

Thanks again!

PR #90, opened 2018-07-01
Add createSubspaceProvider, mirroring Redux's createProvider.

### LINKED ISSUE #89: Support alternate store keys.
Redux-subspace should configuration for attaching to alternate redux store keys.

This will support usage in situations where there may actually be more than one Redux store, though that is something that this library will hopefully remove the need for.

Also it will support situations like mine, where I am building a set of reusable, redux-bound wrapper components which return rendering control to their parent context. In this case in order for the application to function properly my subspace for the wrappers needs to have its own store key, because using  SubspaceProvider on the `store` context key would prevent the wrapped application components from accessing the non-subspaced store.

conartist6:
| Q                       | A 
| Fixed Issues?           | Fixes #89
| Minor: New Feature?     |  👍

See changed documentation for a description of the changes : )


mpeyper:
Nice work on this.  I've left a few comments for discussion, but overall happy with this approach.

As a side note for your intended use case, [the way react-redux uses context is changing soon](https://github.com/reduxjs/react-redux/issues/950) so it might not continue to work going forward, but it's too soon to tell on any of that, so just a heads up.

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -10,37 +10,41 @@ import React, { Children } from 'react'
 import PropTypes from 'prop-types'
 import { subspace }  from 'redux-subspace'
 
-class SubspaceProvider extends React.PureComponent {
+export function createSubspaceProvider(storeKey = 'store', parentStoreKey = 'store') {
```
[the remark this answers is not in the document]
conartist6: Arrow function, sure. I considered doing that actually. My opinion is that if you want that behavior the argument order should be swapped:
``` 
export const createSubspaceProvider = (parentStoreKey = 'store', storeKey = parentStoreKey, ) => {
```
Actually though any way you slice it someone is going to have to specify a default parameter they don't really care about in order to specify one that they do. Maybe a literate syntax like  `createProvider().withStoreKey(key)` would be best of all?

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -10,37 +10,41 @@ import React, { Children } from 'react'
...
-}
+    SubspaceProvider.contextTypes = {
+        [parentStoreKey]: PropTypes.object.isRequired
+    }
 
-SubspaceProvider.contextTypes = {
-    store: PropTypes.object.isRequired
-}
+    SubspaceProvider.childContextTypes = {
+        [storeKey]: PropTypes.object
```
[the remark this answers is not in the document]
conartist6: Haha ok I'll find out.

conartist6:
Yeah I've read though the threads on context. I think this should be even easier to do once we're just passing around {Provider, Consumer} pairs.

packages/react-redux-subspace/src/components/SubspaceProvider.js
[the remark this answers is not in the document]
mpeyper: I think I agree about the defaulting order.  If the parent store key is changed, you probably want the store key to be changed too.

The scenarios I see:

1. parent store default, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider()
   ```
2. parent store override, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider('other')
   ```
3. parent store default, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider('store', 'other')
   ```
4. parent store override, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider('other', 'store')
   ```
5. parent store override, child store different override
   ```js
   const SubspaceProvider = createSubspaceProvider('other1', 'other2')
   ```

As you said, in both 3. and 4. one has be be defined as the default key, which kinda sucks.

How about
```js
export const createSubspaceProvider = ({ storeKey = 'store', parentStoreKey = storeKey, childStoreKey = storeKey } = {}) => {
```

1. parent store default, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider()
   ```
2. parent store override, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider({ storeKey: 'other' })
   ```
3. parent store default, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider({ childStoreKey: 'other' })
   ```
4. parent store override, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider({ parentStoreKey: 'other' })
   ```
5. parent store override, child store different override
   ```js
   const SubspaceProvider = createSubspaceProvider({ parentStoreKey: 'other1', childStoreKey: 'other2' })
   ```

With this, you'd only specify the one you want to override, with `storeKey` allowing both to be overridden. 

Or, possibly, accepting either a string or an object.  If a string is provider, it sets both `parentStoreKey` and `childStoreKey` to be the string, something like:

```js
const DEFAULT_STORE_KEY = 'store'

const createSubspaceProvider = (storeKey = DEFAULT_STORE_KEY) => {
  const { parentStoreKey = DEFAULT_STORE_KEY, childStoreKey = DEFAULT_STORE_KEY } = typeof(storeKey) === 'string' 
    ? { parentStoreKey: storeKey, childStoreKey: storeKey } 
    : storeKey

  // ...
}

1. parent store default, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider()
   ```
2. parent store override, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider('other')
   ```
3. parent store default, child store override
   ```js
   const SubspaceProvider = createSubspaceProvider({ childStoreKey: 'other' })
   ```
4. parent store override, child store default
   ```js
   const SubspaceProvider = createSubspaceProvider({ parentStoreKey: 'other' })
   ```
5. parent store override, child store different override
   ```js
   const SubspaceProvider = createSubspaceProvider({ parentStoreKey: 'other1', childStoreKey: 'other2' })
   ```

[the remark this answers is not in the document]
Comment (conartist6) [reply packages/react-redux-subspace/src/components/SubspaceProvider.js]: Yeah I think your last suggestion hits the nail on the head. It's verbose to do the thing that it should be difficult/rare to need to do, and when you're just trying to proxy on a setting that you already configured using Redux `createProvider` the syntax is actually exactly the same as createProvider's.

Perfect.

**Comment by conartist6** [comment]
OK, so the `[contextKey]Provider` context key is used for ReactRedux Subscriptions which help ensure that parents components have their updates handled before children. I have updated the API and incorporated that context key into the changeset.

File: packages/react-redux-subspace/src/components/SubspaceProvider.js [hunk 3]
```diff
@@ -10,37 +10,50 @@ import React, { Children } from 'react'
 import PropTypes from 'prop-types'
 import { subspace }  from 'redux-subspace'
 
-class SubspaceProvider extends React.PureComponent {
-
-    getChildContext() {
-        const makeSubspaceDecorator = (props) => props.subspaceDecorator || subspace(props.mapState, props.namespace)
+export function createSubspaceProvider(paramStoreKey = 'store') {
+    let storeKey = typeof paramStoreKey === 'object' ? paramStoreKey.storeKey || 'store' : paramStoreKey;
```
mpeyper: I think I would much prefer to be more explicit about what each `storeKey` is when supplying an object to prevent any confusion.

```js
const DEFAULT_STORE_KEY = 'store'

export const createSubspaceProvider = (storeKey = DEFAULT_STORE_KEY) => {
  const childStoreKey = typeof storeKey === 'object' ? storeKey.childStoreKey || DEFAULT_STORE_KEY : storeKey
  const parentStoreKey = typeof storeKey === 'object' ? storeKey.parentStoreKey || DEFAULT_STORE_KEY : storeKey;

  class SubspaceProvider extends React.PureComponent {
    getChildContext() {
      const makeSubspaceDecorator = (props) => props.subspaceDecorator || subspace(props.mapState, props.namespace)

      return {
        [childStoreKey]: makeSubspaceDecorator(this.props)(this.context[parentStoreKey]),
        [`${childStoreKey}Subscription`]: this.context[`${parentStoreKey}Subscription`],
      }
    }

    render() {
      return Children.only(this.props.children)
    }
}
```

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -10,37 +10,50 @@ import React, { Children } from 'react'
...
+    let parentStoreKey = typeof paramStoreKey === 'object' ? paramStoreKey.parentStoreKey || 'store' : paramStoreKey;
+
+    class SubspaceProvider extends React.PureComponent {
+        getChildContext() {
+            const makeSubspaceDecorator = (props) => props.subspaceDecorator || subspace(props.mapState, props.namespace)
+
+            if (storeKey !== 'store') {
+                console.log('parentStoreKey: ', parentStoreKey)
+                console.log(this.context[parentStoreKey])
```
mpeyper: Did you mean to include these?

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -10,37 +10,50 @@ import React, { Children } from 'react'
...
+            }
+            return {
+                [storeKey]: makeSubspaceDecorator(this.props)(this.context[parentStoreKey]),
+                [`${storeKey}Subscription`]: this.context[`${parentStoreKey}Subscription`],
```
mpeyper: I'm starting to wonder if it would be easier to just use `react-redux`'s `Provider` to inject our subspaced store back into the context instead of trying to replicate how `Provider` does it.  It might make us more resilient to any changes they may make in the future.  

I won't insist on changing it to do that as part of this PR as I don't anticipate them changing anytime soon (other than the huge context API changes but we'll likely have to overhaul a bunch of stuff when that happens anyway).

packages/react-redux-subspace/test/components/SubspaceProvider-spec.js
```diff
@@ -37,6 +37,74 @@ describe('SubspaceProvider Tests', () => {
...
+        let testComponent = render(
+            <Provider store={mockStore}>
+                <AlternateContextKeySubspaceProvider mapState={state => state.subState}>
+                    <AlternateContextKeyTestComponent />
+                </AlternateContextKeySubspaceProvider>
+            </Provider>
+        )
+
+        expect(testComponent.text()).to.equal("expected")
+    })
```
mpeyper: wouldn't mind also seeing some cases for the defaulting of the keys when only one is provided, i.e.

* `createSubspaceProvider({ childStoreKey })`
* `createSubspaceProvider({ parentStoreKey })`

packages/react-redux-subspace/src/components/SubspaceProvider.js
conartist6: oops, nope...

mpeyper:
I'm going to close this due to inactivity and the react-redux v6 changes proposed in #103 will make essentially make changing the store key in the legacy context obsolete.

PR #93, opened 2018-07-27
Update docs/example for redux-observable 

evertbouw:
| Q                       | A 
| Documentation only PR   | 👍

Figured I should update the docs while it's still fresh in memory. Might be a bit too soon as the version numbers selected don't exist yet 😃


mpeyper:
Hey @evertbouw.  Thanks again.

It seems like the build is failing on this change, but it seems completely unrelated.  I'll try and work out what is going on, but in the mean time, I'll leave this out (on the off chance it is related).

evertbouw:
I think it was me causing the build fail, since I used version 3 of subspace in the example package.json 😃

Issue #95, opened 2018-08-31
Incompatability with Redux 4 in react-redux-subspace/redux-subspace


carl-berg:
I am trying to used `subspaced` from the `react-redux-subspace`  and am currently getting an error

```
...my-project-path/node_modules/redux-subspace/src/index.d.ts
(34,21): Namespace '"...my-project-path/node_modules/redux/index"' has no exported member 'GenericStoreEnhancer'.
```

```ts
<Route
  path="something"
  component={subspaced((state) => state["module-namespace"], "module-namespace")(Component)}/>
```
### My Setup
| package        | version(s)
| ---------------| -------
| redux          |  4.0.0 |
| redux-subspace | 2.5.0 |
| react-redux-subspace | 2.5.0 |

Not sure if this is already known and part of #87 or not, but if you are aware you can feel free to just close this issue

mpeyper:
Yes, I was aware the types are not compatible, but I do not believe there was ever an issue to capture it.

They have been fixed in master, as part of the redux-observable v1 compatability changes , but have not been released yet as not all packages have been confirmed to be compatible with redux 4 yet.

I'll keep this open until a new version goes out, just in case anyone else comes looking.

mpeyper:
Fixed in #97

PR #97, opened 2018-10-26
Remaining redux v4 compatibility changes

### LINKED ISSUE #87: Redux v4 compatibility
[Redux v4 has been officially released now](https://github.com/reactjs/redux/releases/tag/v4.0.0) so we need to ensure that all of our packages are still compatible.

From my understanding, there aren't many breaking changes so it should hopefully just be a case of bumping the version and running the tests to ensure it all still works.

The `redux` peer dependency should continue to support `^3.0.0` so it will need changing to `^3.0.0 || ^4.0.0` (or however else it's possible to define it that allows consumers to user v3 or v4 without getting a warning).

mpeyper:
This is mostly just bumping versions but there are a few notable exceptions:
1. added the same error during middleware setup that `redux` now throws (one of the reasons `redux` v4 was a breaking change)
2. `redux-loop` v4 upgrade - previously we supported v2 and v3, but in this PR I have dropped v2 support completely.  This is because v4 no longer exports the deprecated v2 functions.  I have retained v3 compatibility, but I'd be open to dropping that too as I don't think upgrading from 3 -> 4 is much effort.
3. real world example upgraded to use`react-router` v4 and `connected-react-router` instead of `react-router-redux`.
4. `redux-observable` example - the redux-observable example was not updated when the v1 upgrade was done so I converted it now (better late than never).
5. `babel` v7 upgrade

I'll comment the specific pieces in the files separately.  

Once merged.  I'd like to do a redux-subspace v3 prerelease for all the packages to start getting feedback on anything else breaking.

Fixes #87, #95


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![npm version](https://img.shields.io/npm/v/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
 [![npm downloads](https://img.shields.io/npm/dm/redux-subspace.svg?style=flat-square)](https://www.npmjs.com/package/redux-subspace)
-[![License: MIT](https://img.shields.io/npm/l/redux-subspace.svg?style=flat-square)](/LICENSE.md)
+[![License: BSD-3-Clause](https://img.shields.io/npm/l/redux-subspace.svg?style=flat-square)](/LICENSE.md)
```
mpeyper: I noticed the `alt` text was incorrect on all our license badges... whoops.

examples/redux-loop/redux-loop-example/package.json
```diff
@@ -3,23 +3,29 @@
...
+    "redux-loop": "^4.4.1",
     "redux-subspace-loop": "^2.5.0"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "eject": "react-scripts eject",
-    "test": "echo 'No Tests'"
-  }
+    "prestart": "npm-link-shared ./node_modules/redux-subspace-loop/node_modules . redux-loop",
```
mpeyper: at some point during redux-loop v3, they started using a`Symbol` for marking if a reducer returned a loop.  The issue with that for us is that `Symbol` created in one file is not the same reference as an identical `Symbol` created in another and because lerna symlinks `redux-subspace-loop` into the example, it has it's own copy of redux-loop so the `Symbol`s are no longer the same so loops started to no look like loops anymore.  This hack just symlinks the version or `redux-loop` from `redux-subspace-loop` into the example so they are the same file again.

examples/redux-loop/redux-loop-example/src/counter.js
```diff
@@ -148,7 +148,7 @@ export const reducer = createReducer({
   [Actions.incrementBothStart]: (state, amount) => {
     console.log('both start');
     return loop(state,
-      Cmd.batch([
+      Cmd.list([
```
mpeyper: A far as I can tell, this is the extent of redux-loop v3->v4 changes.

package.json
```diff
@@ -10,12 +10,10 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap --concurrency=1",
```
mpeyper: new version of lerna seems to barf every couple of builds with an error about a callback not being called.  Setting concurrency to 1 slows it down but is more stable.

mpeyper:
@jpeyper I'm also not sure what state this leaves #96

mpeyper:
oh, and it looks like some of this was already covered in #93

PR #98, opened 2018-12-07
Inside a Component wrapped by SubspaceProvider we can now call actions which correspond to different namespace

prasadyash2411:
| Q                       | A 
| Fixed Issues?           | Fixes `#1`, Fixes `#2` 
| Documentation only PR   | 
| Patch: Bug Fix?         | 
| Minor: New Feature?     |  👍
| Major: Breaking Change? | 

So the problem which i was solving using redux subspace was that in a single page i had multiple data graphs. The number of graphs could range from anywhere between 1 to 20 data  graphs. All these dat graphs were required to hit the same api and derive their data and as a result their was only one type of action being trigerred and only one reducer. But now since more than one graph with different datas were to be shown , I had to fire multiple actions(one for each of the graph). As a result of this the reducer used to get updated with the data of the last trigerred action and all the graph used to show the same data.

This problem was solved by namespacing all the epics and reducer with respect to the unique ids of the graph and by enclosing the graph container inside the subspace provider and passing id of each graph in the namespace prop so that all the graphs fire actions that are namespaced with respect to their id and as a result all the data get mapped to their corresponding namespaced reducer. This solved the problem of each graph showing their respective data which come on firing their respectice actions.

But the problem that i came across was that inside the graph container i could only fire actions that correspond to their respective id as the HOC subspace provider was namespacing all the actions with the respect to the value that was being passed in the namespace prop. So if I was to say reload the second graph(id-2) from the first graph (id-1) , I was supposed to fire an action that was namepaced with id-2  and i could only do this inside my graph  container because i was getting all ids inside it. I could fire namespaceAction with id-2 but the HOC subspace provider was again namespacing the action and as a result i was getting an action like this id1/id2/action but i only wanted id2/action.
I tried to do this with the action ---> globalAction(namespaceAction('id2')(action)) but it still didnt work.

The Solution :
I forked the redux subspace repo and added a new function called overRideNameSpaceAction and it is very similar to namespacedAction. Just that when I call it makes a flag called alreadyNameSpaced --> true and I added a check in namespaceAction function that if the flag AlreadyNamespaced is true then dont namespace it again. This prevents the action inside the SubSpaceprovider  which has been enclosed in overridenamespaceAction to not get namespaced again.

So now inside the graph container when i am inside graph with id-1, inside which all action get namespaced with id-1 , I can simply call overRideNameSpacedAction(id2)(action) and i can prevent subspaceprovidder from namespacing it agian.
So now i can reload graph with id-2 from graph container with id-1.

Change I made to the repo
1)Added a new file overRideNameSpaceAction
2) Added one more condition to namespaceAction caled alreadyNamespaced
3)removed lib from gitignore to build the file and push to master
4)Changed package.json of react-redux-subpace  so that in dependency of redux-subspace it points to the redux subspace package in which i have made changes
5)Change path of react-redux-subspace to redux-subspace/packages/react-redux-subspace as the npm module was not being able to locate


mpeyper:
> I tried to do this with the action ---> globalAction(namespaceAction('id2')(action)) but it still didnt work.

My initial thought was to just do this and I wondered why it didn't _just work_.  I even made [a sandbox to prove it worked](https://codesandbox.io/s/j735pxlqx5) (_**Narrator:** It didn't_).

The reason this doesn't work as initially expected is because the current behaviour for a `globalAction` is to not modify the `type` at all when passing it into the reducers, so although the action has the expected namespace when it get to the root store (i.e. `id2/action`) if retains that namespace when it reaches the reducer so it does not update the state as expected.

Given this peculiarity of the `globalAction` method, I think this is not a bad approach to take if we want to support this use case.

The reason I suggest that we may not want to support it is that, at it's core, `redux-subspace` is about isolation of the subspaces from one another.  The `globalAction` helper is provided as an escape hatch for communication between subspaces, but it comes with the caveat of requiring the action to be truely global.  On the other side there `namespacedAction` which was intended to allow parent subspaces to push actions into their children, for which they know and control the namespaces.  By opening up the concept of subspace to subspace communication, there is an implicit understanding that the subspaces are aware of each other.  This may be somewhat acceptable when dealing with sibling subspaces, but gets very messy when nested subspaces are considered.

For example, consider an app with a subspace structure like:

* root
  * parent1
    * child1
  * parent2
    * child2

If `child1` dispatches an `overrideNameSpaceAction('parent2/child2')(action)` then it not only requires knowledge of the existence of `child2` but also the structure of the app.  If `parent2` ever gets renamed, or moves deeper into the structure, or if `child2` gets renamed or moved higher into the structure, then `child1` silently breaks.

But theory aside, I'm happy to explore this a bit further as I'm well aware that often the requirements of the app often call for us to just _make it work_.

My first thought is I wonder if this could be achieved through a [custom middleware](https://ioof-holdings.github.io/redux-subspace/docs/advanced/middleware/CustomMiddleware.html).  Something like:

```js
import { compose } from 'redux'
import { globalAction, namespacedAction, applyToRoot } from 'redux-subspace'

const actionWrapper = namespace => {
  return compose(
    globalAction,
    namespacedAction(namespace),
    action => ({ ...action, absoluteAction: true })
  )
}

export const absoluteAction = namespace => action => actionWrapper(namespace)(action)

export const absoluteActions = applyToRoot(store => next => action => {
  if (action.absoluteAction) {
    return store.dispatch({ ...action, globalAction: false, absoluteAction: false })
  }
  return next(action)
})
```

[Here is sandbox of it working](https://codesandbox.io/s/186y96orwq).

The idea here is that a action wrapper puts a known flag on the action, and then a middleware that only gets applied to the root store looks for it, then strips reverses the global and absolute action flags so it gets treated as a standard action when it gets to the reducers.

Ideally, `redux-subspace` would provide the ability to un-`globalAction` the action (`localAction`?), so that this could be implemented without knowing any of the internals of the library.

This is more or less the same idea as what you have in the PR, but is completely implement in user space, so we don't need to pollute the internals of `redux-subspace` to support it.

As a personal preference, I think I prefer the `absolute...` to `overridden...` naming but I'm open to hear thoughts for and against either naming (or alternative suggestions).

This solution could also be published to npm as a stand alone package, or just kept locally in your project.  We may also consider including it (or a variant of it) in the core package, but I won't make any promises to do so at this stage.

How does this sound?

mpeyper:
I'm going to close this due to inactivity and my proposed solution that does not require any changes to this library.  If anyone wants to discuss this further, please open an issue and we'll talk about it more.

PR #100, opened 2018-12-11
add defaultState from the original reducers to the namespaced ones

### LINKED ISSUE #99: Using a prefilled state with namespaced reducers causes loss of the default store values.
Hello!
I faced some issues when I was trying to use namespaced reducers while creating the redux store with prefilled data.

I've made a simple [repo](https://github.com/NikitaKolokoltsev/redux-subspace-preloadedState) to illustrate the problem.

Briefly, the bug is appearing when you have some reducer, lets say `viewReducer`:
```javascript
let defaultState = {
  design: 'flat',
  detalizationLevel: 'small'
};

export default function viewReducer(state, action) {
  state = {
    ...defaultState,
    ...state
  }

  ...
```
Which is appearing as a part of the `settingsReducer`:
```javascript
const settingsReducer = namespaced('settings')(
  combineReducers({
    view: viewReducer,
    ...another reducers belonging to the settings section...
  })
);

const appReducer = combineReducers({
  settings: settingsReducer,
  ...there can be other reducers for other parts of the application...
})
```
So right now the store hierarchy is something like that:
```javascript
{
  settings: {
    view: {
      design: 'flat',
      detalizationLevel: 'small'
    },
    ...
  },
  ...
}
```
**So the issue itself:** If you try to create redux store with prefilled state:
```javascript
const prefilledState = {
  settings: {
    view: {
      design: 'material'
    }
  }
}

const store = createStore(appReducer, prefilledState);
```
Then it will result into the store with missing default values:
```javascript
store.getState();
# => {
       settings: {
         view: {
           design: 'material'
         }
       }
     }
```
As you can see the `detalizationLevel` field is not present.

### Possible solution
This issue exists because of the `processAction()` function's code.
According to the code of the redux's `createStore()` - *redux/src/createStore.js* function:
```javascript
// When a store is created, an "INIT" action is dispatched so that every
// reducer returns their initial state. This effectively populates
// the initial state tree.
dispatch({ type: ActionTypes.INIT })
```
This action is a special unique redux-specific action which type starts with `@@redux/...`.
This will trigger all reducers because it wouldn't find appropriate `case` for this action and it will initialze store with default values which are specified in the reducers.
But according to the code of `processAction()` - *redux-subspace/src/actions/processAction.js* function:
```javascript
import hasNamespace from '../actions/hasNamespace'
import isGlobal from '../actions/isGlobal'

const processAction = (namespace) => (action, callback, defaultValue) => {
    if (!namespace || isGlobal(action)) {
        return callback(action)
    } else if (hasNamespace(action, namespace)) {
        return callback({...action, type: action.type.substring(namespace.length + 1)})
    } else {
        return defaultValue
    }
}

export default processAction
```
As this redux-specific actions are not namespaced, first two conditions in `processAction()` function will be false so all namespaced reducers will return **ONLY(!!!)** provided prefilledState and that will lead to some missing default values untill the user duplicates them in the prefilledState.

So I made a quick fix for it in the corresponding PR.
But I am afraid a bit of the fact that if it was written like so, then there were a reasons for it. So, please, tell me if I am missing something.

### Context
Situation with prefilledState is quite common when using server side rendering with React, when you fetched data from the API and need to generate the store from which yout React-application will take the data to render.

### Your Setup

| package        | version(s)
| ---------------| -------
| redux          |  4.0.1
| redux-subspace |  3.0.0

Thank you guys for your really helpfull library!

NikitaKolokoltsev:
| Q                       | A 
| Fixed Issues?           | Fixes #99
| Patch: Bug Fix?         |  👍

Sorry for the mess in commits :)


packages/redux-subspace/src/actions/reduxSpecific.js
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const REDUX_PREFIX = '@@redux/'
+
+const reduxSpecific = (action) => action.type.startsWith(REDUX_PREFIX)
+
+export default reduxSpecific
```
taleh007: add correct file ending

packages/redux-subspace/src/actions/reduxSpecific.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright 2017, IOOF Holdings Limited.
```
taleh007:
```suggestion
 * Copyright 2018, IOOF Holdings Limited.
```

mpeyper: Github wont let me commit this, but yes.

packages/redux-subspace/src/actions/processAction.js
```diff
@@ -8,9 +8,10 @@
 
 import hasNamespace from '../actions/hasNamespace'
 import isGlobal from '../actions/isGlobal'
+import isReduxSpecific from '../actions/reduxSpecific'
 
 const processAction = (namespace) => (action, callback, defaultValue) => {
-    if (!namespace || isGlobal(action)) {
+    if (!namespace || isGlobal(action) || isReduxSpecific(action)) {
```
mpeyper: I'd be tempted to move this into the `isGlobal` check.  Essentially we are saying that redux actions are _always_ global actions.

mpeyper: This case should get a specific unit test.

NikitaKolokoltsev: I'll add tests a bit later

mpeyper:
Oh, I forgot to do the usual preamble...

Thanks for taking the time to improve `redux-subspace`. Don't forget to [include yourself as a contributor](https://github.com/ioof-holdings/redux-subspace/blob/master/CONTRIBUTING.md#add-yourself-as-a-c) before you're done!

mpeyper:
Thanks for this. I'll put together a release later today.

PR #103, opened 2019-02-20
Updated react-redux bindings to v6 and updated examples

### LINKED ISSUE #102: TypeError: undefined is not an object (evaluating 'store.subspaceOptions')
### Is it a bug, feature request or question?

It seems a bug, I always get error below when I try the sample code provided in the documentation.
```
TypeError: undefined is not an object (evaluating 'store.subspaceOptions')
```
### Which package(s) does this involve?

react-redux-subspace

### Input Code

I created a blank react-native app using Expo, then the `App.js` files has the following content:

```js
import React from 'react';
import { createStore, combineReducers } from 'redux';
import { namespaced } from 'redux-subspace';
import { Provider } from 'react-redux';
import { SubspaceProvider } from 'react-redux-subspace'
import { StyleSheet, Text, View } from 'react-native';

// dummy reducer
const foo = (state = 42) => state

const rootReducer = combineReducers({
  foo: namespaced('foo')(foo),
})

const store = createStore(rootReducer)

const Foo = () => (
  <View style={styles.container}>
    <Text>Open up App.js to start working on your app!</Text>
  </View>
)

export default () => (
  <Provider store={store}>
    <SubspaceProvider mapState={state => state.foo} namespace="foo">
      <Foo />
    </SubspaceProvider>
  </Provider>
);

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

### Expected Behavior

I should see a screen with the text `Open up App.js to start working on your app!`

### Current Behavior

I get the following error and stacktrace:
```
[18:16:27] Warning: Failed context type: The context `store` is marked as required in `SubspaceProvider`, but its value is `undefined`.
- node_modules/prop-types/checkPropTypes.js:20:20 in printWarning
- ... 22 more stack frames from framework internals

[18:16:31] TypeError: undefined is not an object (evaluating 'store.subspaceOptions')

This error is located at:
    in SubspaceProvider (at App.js:28)
    in Provider (at App.js:27)
    in _default (at withExpoRoot.js:22)
    in RootErrorBoundary (at withExpoRoot.js:21)
    in ExpoRootComponent (at renderApplication.js:34)
    in RCTView (at View.js:44)
    in RCTView (at View.js:44)
    in AppContainer (at renderApplication.js:33)
- node_modules/redux-subspace/lib/index.js:377:69 in <unknown>
- node_modules/react-redux-subspace/lib/index.js:37:47 in getChildContext
- ... 21 more stack frames from framework internals
```

If I remove the `SubspaceProvider` then the app renders, but of course I don't get the subspace.

### Your Setup

Here are my dependencies from `package.json`:
```json
 "dependencies": {
    "expo": "^32.0.0",
    "react": "16.5.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
    "react-redux": "^6.0.0",
    "react-redux-subspace": "^3.0.1",
    "redux": "^4.0.1",
    "redux-subspace": "^3.0.1"
  },
```

mpeyper:
| Q                       | A 
| Fixed Issues?           | Fixes #102
| Documentation only PR   | 
| Patch: Bug Fix?         | 
| Minor: New Feature?     | 👍 
| Major: Breaking Change? | 
| Tests Added + Pass?     | no new tests, existing tests pass

Support for react-redux v6.x.x using the new context API.

More discussion in the code.


examples/real-world/src/root/routes.js
```diff
@@ -1,22 +1,26 @@
 import React from 'react'
 import { Route, Switch } from 'react-router' 
 import { SubspaceProvider } from 'react-redux-subspace'
+
+// work around for lerna/npm link issue for local development
+import { ReactReduxContext } from 'react-redux'
```
mpeyper: You'll see this a lot in the review.  When `lerna` links the packages using `npm link` they end up with different instances of `react-redux` so the `React.createContext(...)` calls create a different context so `react-redux-subspace` cannot access the store from the examples `Provider`.  This follows the "BYO context` concept from `react-redux` but provides the actual context instance from the example's version.

This works, but it's a bit gross and not really how you would use in a real app, which makes having it in all the examples a bit annoying.

packages/react-redux-subspace/package.json
```diff
@@ -31,47 +31,49 @@
     "url": "https://github.com/ioof-holdings/redux-subspace.git"
   },
   "dependencies": {
-    "@types/react": "^16.4.18",
-    "hoist-non-react-statics": "^3.0.1",
-    "prop-types": "^15.6.0",
+    "@types/react": "^16.8.3",
+    "hoist-non-react-statics": "^3.3.0",
+    "memoize-one": "^5.0.0",
```
mpeyper: new dependency

packages/react-redux-subspace/package.json
```diff
@@ -31,47 +31,49 @@
...
+    "prop-types": "^15.7.2",
     "recompose": "^0.30.0",
     "redux-subspace": "^3.0.1"
   },
   "peerDependencies": {
     "react": "^15.1 || ^16.0",
-    "react-redux": "^5.0.0",
+    "react-redux": "^5.0.0 || ^6.0.0",
```
mpeyper: current implementation should be backwards compatible, but it has complicated the internals.  We could do a clean cut over and do a major bump instead.  I'm just worried about how we currently distribute `subspace` and `dynostore` through our bundles, not having backwards compatibility will make upgrading harder.

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -25,22 +58,40 @@ class SubspaceProvider extends React.PureComponent {
...
-
-SubspaceProvider.contextTypes = {
-    store: PropTypes.object.isRequired
+    subspaceDecorator: PropTypes.func
 }
 
-SubspaceProvider.childContextTypes = {
-    store: PropTypes.object
-}
+const storeAccess = compose(contextApi, legacyContext)
```
mpeyper: 2 way to access the store, both pass the store as a prop to the underlying `SubspaceProvider` only if they find it in their respective methods.  Technically with this change, a store can be passed directly as a prop if there is no `Provider` (or `SubspaceProvider`) higher in the tree.

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -25,22 +58,40 @@ class SubspaceProvider extends React.PureComponent {
     }
 }
 
+SubspaceLegacyContextOutput.childContextTypes = {
+    store: PropTypes.object
+}
+
+class SubspaceProvider extends React.Component {
+    makeSubspace = memoizeOne((mapState, namespace, subspaceDecorator, store) => {
```
mpeyper: We actually get a perf boost from this change as we are only going to recalculate the subspace if the params to create it change.  Previously we would recalculate it any time `getChildContext` was called (which I believe is on every render).

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -25,22 +58,40 @@ class SubspaceProvider extends React.PureComponent {
...
+        const createSubspace = subspaceDecorator || subspace(mapState, namespace)
+        return createSubspace(store)
+    })
+
+    render() {
+        const { mapState, namespace, subspaceDecorator, store, children } = this.props
+        const subspace = this.makeSubspace(mapState, namespace, subspaceDecorator, store)
+        return (
+            <Provider key={subspace.namespace} store={subspace}>
+                <SubspaceLegacyContextOutput subspace={subspace}>
```
mpeyper: This will cause the subspace to always be put into legacy context for and downstream `SubspaceProviders` or `dynostore` `dynamic` components using older versions to access.

packages/react-redux-subspace/package.json
jpeyper: Yeah we will definately want to have this version supporting both.

Even if we release the major bump soon after just to cleanup and prepare for a "no legacy context" future, well probably be running this version for a while until we can port all the micro-frontends over.

packages/react-redux-subspace/src/components/SubspaceProvider.js
```diff
@@ -25,22 +58,40 @@ class SubspaceProvider extends React.PureComponent {
...
+    subspaceDecorator: PropTypes.func
 }
 
-SubspaceProvider.childContextTypes = {
-    store: PropTypes.object
-}
+const storeAccess = compose(contextApi, legacyContext)
 
-export default SubspaceProvider
+export default storeAccess(SubspaceProvider)
```
mpeyper: because this maintains legacy context support, subspacing a single component using

```js
<SubspaceProvider mapState={(state) => state.example} namespace='example>
  <Component />
</SubspaceComponent>
```

ends up with a component heirarchy of

```js
<SubspaceContextApi>
  <ReactReduxContext.Consumer>
    <SubspaceLegacyContextInput>
      <SubspaceProvider>
        <Provider>
          <ReactReduxContext.Provider>
            <SubspaceLegacyContextOutput>
              <Component />
            </SubspaceLegacyContextOutput>
          </ReactReduxContext.Provider>
        </Provider>
      </SubspaceComponent>
    </SubspaceLegacyContextInput>
  </ReactReduxContext.Consumer>
</SubspaceContextApi>
```

If you add in the `subspaced` HOC, the top level `Provider` and the layers for the `connect` HOC of the component, this starts to look a bit crazy.  In fairness, the legacy context support is only 2 of the 7 layers, so not having it only makes the problem a little bit better.

examples/real-world/src/root/routes.js
jpeyper: This is one annoying thing about reacts new context api.

Could we have a post install script that removes the nested react-redux packages in node_modules so they all resolve to the same one?

At least then the fix is hidden away in the package.json and not in the example code.

packages/react-redux-subspace/src/components/SubspaceProvider.js
jpeyper: Providing your own store could be useful for testing?

packages/react-redux-subspace/src/components/SubspaceProvider.js
jpeyper: Definately a necessary evil.

examples/real-world/src/root/routes.js
jpeyper: Actually, that would break the build of the actual package... Ignore that.

mpeyper: The issue with that is the file is not in the same directory tree, so the node package resolution does not find the package at all. I have done a trick in the past to `npm link` the troublesome package back to the other's as a post-installation task, but I found it to be slow and required a bit of hacking about to get working.  At least this is actual functionality offered by both `react-redux` and now `redux-subspace`, it just sucks a bit that it's the defacto example.

packages/react-redux-subspace/package.json
jpeyper: Talking internal to ioof now. Because of the new context api and having to be the same reference, we're going to externalize subspace (and dynostore?), Upgrade all the micro-frontends so they are using the same version from the parent, then upgrade react-redux.

We actually won't need to support both contexts at once? If we first externalize the old version and then cut over to the new.

We may still want to have this version for others to migrate with.

examples/real-world/src/root/routes.js
jpeyper: At least it won't break is hey do do it in production code (it'll be the same as the default value)

Maybe adjust the comment a bit to make it more obvious?

"// Note: please don't copy this line line (and associated prop usage) to your project, you are unlikely to need to do this for standard subspace usage.
// This is caused by Leena using  non link..."

packages/react-redux-subspace/src/components/SubspaceProvider.js
jpeyper: Anyone running an app complex enough to need subspace likely has worse demons in their component tree.

packages/react-redux-subspace/package.json
mpeyper: Because the context we use comes from `react-redux` which is already externalised, we don't _have_ to externalise subspace or dynostore for this change...

But I **_really, really_** think we should so that future changes like this are easier.  Anything where there is a provider in one bundle and a consumer in another should, in my opinion, be externalised.

examples/real-world/src/root/routes.js
mpeyper: lol at autocorrect, but I like the idea.

jpeyper: That'll teach me for reviewing on my phone.

erdembas:
How long will I wait for the new version? Will you accept the pull request? Looking forward to the new version. @jpeyper

mpeyper:
Hi @erdembas,

In all honesty, I've been busy at work and as we aren't in a position to actually update react-redux to v6 ourselves, this has taken a bit of a backseat in my priorities.

We still need to:
1. decide if we want to support backwards compatibility for legacy context
2. ensure there is a safe migration path for parent/child subspaces

I'll try to get this closed off soon.  Sorry for the inconvenience.

erdembas:
Hi @mpeyper ,

I will be very pleased if it is closed soon. I'm waiting :)

raybooysen:
As another bit of info, V7 is already out for Redux

raybooysen:
To work around this, i've created two components that let you shim between redux-subspace (in it's current form) and redux 7.x:

```
import {Store} from 'redux';
import React from 'react';
import PropTypes from 'prop-types';
import {Provider, ReactReduxContext} from 'react-redux';

interface InnerShimProps {
    store: Store;
}
class InnerShim extends React.Component<InnerShimProps> {
    static childContextTypes = {
        store: PropTypes.object
    };

    getChildContext() {
        return {store: this.props.store};
    }

    render() {
        return this.props.children;
    }
}

export const SubspaceStoreShim: React.FC<{}> = props => {
    const Context = ReactReduxContext;
    return (
        <Context.Consumer>
            {value => {
                return <InnerShim store={value.store}>{props.children}</InnerShim>
            }}
        </Context.Consumer>
    );
};


export class SubspaceToReduxShim extends React.Component<{}> {
    render() {
        return (
            <Provider store={this.context.store}>
                {this.props.children}
            </Provider>
        );
    }

    static contextTypes = {
        store: PropTypes.object
    }
}
```

`SubspaceStoreShim` takes the redux store from the new context API and switches it to the legacy API so that `SubspaceProvider` can consume it.

`SubspaceToReduxShim` can then be used AFTER the SubspaceProvider has been used so subsequent redux `connect` calls can access the store again.

mpeyper:
I know people have been waiting for this, and I'm sorry it's dragging on.  I plan on revisiting this tomorrow and getting it updated to the [Redux 7.1 release](https://github.com/reduxjs/react-redux/releases/tag/v7.1.0).

mpeyper:
Closing this now that #117 is up.

PR #106, opened 2019-05-23
Introduce Parent Spaces

### LINKED ISSUE #105: Any ideas on subspacing components with render props?
I have certain components that have render props. Here's an example:
```js
<SubspaceProvider mapState={state => state.myTable} namespace="MY_TABLE">
  <Table
    rows={[...]}
    renderRow={data => {
      return (
        <TableRow>
          <FooCell foo={data.foo}>
          <BarCell bar={data.bar}>
        </TableRow>
      );
    }}
  />
</SubspaceProvider>
```

My table component is in a subspace, so that it has a reducer which can handle its own selection logic, but those rows are being rendered inside the subspace, and they need to have their parents redux context.

Right now I subspace each row using `subspace((_, rootState) => rootState)`, but this clearly sucks. For one thing it makes the strong assumption that the parent state selector is the root state selector, which is rather limiting. For a second, there's no way to do anything similar with action namespaces.

Do the authors have any thoughts as to how this might be supported, or if it is worth supporting?

conartist6:
Fixes #105 in the manner discussed there.

I think all that's missing is TS type tests. I'll write those, but that's a small thing and there's lots to review already.


mpeyper:
I'll give this a proper review when I get to work, but looking on my phone it looks pretty well done.

My only initial thought is if `parentSpace` needs to exist in the core package or if this problem is somewhat unique the react component tree and render props use case. 

Perhaps not and it applies in other scenarios too? It does open the question up for if we need a `parentSpaced` function for each of the other integration packages as well?  I'm not asking you to add those in this PR, just something for me to think about.

I see some test have failed on the build as well, in case you were not aware.

conartist6:
You are right that perhaps it doesn't need to be in core. We'd still need the hierarchy enhancer obviously but maybe not the unsubspace function. After all if you're using that function you have the store that you just passed in to it...

conartist6:
I removed the addition to the core package for the reasons mentioned above.

packages/redux-subspace/test/enhancers/hierarchyEnhancer-spec.js
```diff
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import rootStoreEnhancer from '../../src/enhancers/rootStoreEnhancer'
+import rootStoreEnhancer from '../../src/enhancers/hierarchyEnhancer'
```
mpeyper: should probably rename in variable to `hierarchyEnhancer` as well to save someone confusion down the track.

conartist6:
A search reveals that there are no more references to `rootStoreEnhancer` now. Anything else that needs to happen with this?

conartist6:
I was missing a typescript test for `<ParentSpaced />`. Just added that.

mpeyper:
Sorry for the delat @conartist6.  There was a failing test in travis, but when I looked into it, the first TS test has been failing on the master branch (an other branches) for months, so can't blame you for that one 😉 

I'll merge this now.  I've got some in progress changes in my checkout at the moment so I'll hit the release button next time it's clean (maybe tomorrow?)

conartist6:
@mpeyper Can you not release this yet, I think I made a mistake : (

conartist6:
I thought I messed something up, but I didn't. Disregard the previous (now deleted) comment.

mpeyper:
This has been published now as [v3.1.0](https://github.com/ioof-holdings/redux-subspace/releases/tag/v3.1.0).  I'm struggling with an error when publishing the docs, but you should know how to use it without referring to the API reference ;)

mpeyper:
Docs are up now.  Thanks again 👍

PR #109, opened 2019-06-14
Fix build

conartist6:
Just trying to fix the build so that I can get my stuff shipped. It works locally, but hopefully not throwing away the error will make it clear why travis is failing.


conartist6:
Made sure that errors weren't discarded, but there were no errors so I just bumped the timeout. I'm guessing since lerna runs a lot of stuff in parallel you need to be a bit more lenient with those timeouts than you might otherwise be.

conartist6:
As a note, the only change required to fix the build is bumping the timeout for the react-redux-subspace typescript tests, but hopefully bumping all the timeouts will make sure that this is unlikely to happen again.

mpeyper:
Hey @conartist6,

Thanks for this. I wasn't holding up the release because if the failing test, it was just my available time (or lack thereof) to actually run the publish steps and write up the release notes.

I'll endeavour to do it sometime this evening, but I'm going out for dinner with my family, so may not be for a few more hours.

conartist6:
No pressure! Just thought I'd clear out one of the obstacles.

PR #137, opened 2019-09-12
Prevent subspace from being recreated if mapState function changes reference

mpeyper:
| Q                       | A 
| Fixed Issues?           | 
| Documentation only PR   | 
| Patch: Bug Fix?         | 
| Minor: New Feature?     | 
| Major: Breaking Change? | 👍

One of my biggest issues with the recent `react-redux` v7 update was that, despite my best effort, if a function was passed for `mapState` that was recreated each `render` cycle, the subspace gets recreated as well.

I know, I know, [`useMemo` and `useCallback` aren't the be-all-and-end-all](https://kentcdodds.com/blog/usememo-and-usecallback) and rerender's aren't necessarily the worst thing, but one instance where is can be critical is the [when the value is used in context](https://kentcdodds.com/blog/how-to-optimize-your-context-value), so it jsut bugged me.

Luckily, I stumbled upon [this tweet by James K Nelson](https://twitter.com/james_k_nelson/status/1171276987144392704) that gave me some inspiration.  The boiled down version of the idea is this:

```js
const useCustomHook = callback => {
  const ref = useRef(callback)
  ref.current = callback
  return useCallback(() => ref.current(), [ref])
}

```

So what's happening here? 

1. The `ref` is initialised to the first version of `callback`
   * Note: the `ref` is only initialised on the first render, it doesn't matter what you pass to `useRef` on subsequent renders as it will just be ignored
   * I think technically this step is not required, but I think it's more clear than just having `const ref = userRef(null)`... maybe?
2. We set the `ref.current` to `callback, mutating the ref to hold the current version of `callback`
   * On subsequent renders this will be how we allow the a newer value to be used
3. We create a memoized callback, wrapping a call to the ref of our provided callback, passing the `ref` in the dependency list
   * In this example, this makes the callback effectively static
   * The callback we use here uses the `ref.current` value internally to always use the current version of the `callback` without requiring a new outer callback to be created each render

In hindsight, this seems pretty straight forward.

One complication I had to work around when applying this solution to `useSubspace` was that the `mapState` parameter can also be provided as a `string` this meant I had to get a bit creative on how we treat the dependencies and use the ref inside the `useMemo`.

Another advantage of this approach is that I was able to remove the `subspaceDecorator` prop that was only used by the `subspaced` HOC as a way of working around the fact that we couldn't reliable memoize the `subspace` function.  This also meant `useSubspaceAdvanced` could be removed and just have `useSubspace` handle it all.

That change means that this is technically (yet another) breaking change.  I'm very tempted to just make it a minor (per is a feature, right?) as `storeDecorator` has never been an advertised or documented part of the public API.  

Speaking of documentation,  I also found 2 errors in our current docs while working on this which have also been included in this PR.


packages/react-redux-subspace/src/hooks/useSubspace.js
```diff
@@ -6,31 +6,40 @@
...
+  const subspacer = useMemo(() => {
+    const mapStateHandler = isMapStateFunction
+      ? (state, rootState) => mapStateRef.current(state, rootState)
+      : mapStateRef.current
+
+    return subspace(mapStateHandler, namespace)
+  }, [
+    isMapStateFunction ? mapStateRef : mapStateRef.current,
+    namespace
+  ])
```
travikk: Isn't that basically the same thing as the following?

```javascript
const isMapStateFunction = typeof mapStateRef.current === "function";
let subspacer;

if (isMapStateFunction) {
   subspacer = useMemo(() => subspace(mapState, namespace), [namespace])
} else {
  subspacer = useMemo(() => subspace(mapState, namespace), [
    mapState,
    namespace
  ]);
}
```

In case we're passing a function, compute the subspacer only when the namespace changes, otherwise compute it when the mapState object or namespace changes.

mpeyper: Not quite.  This would fail to allow `mapState` function from being changed if the `namespace` did not.

I get what you're saying generally with using `isMapStateFunction` once and switching out which `useMemo` get's used.  The main reason I have a single `useMemo` was that I was trying hard not to violate the [one of the rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).  I know that only a single `useMemo` would be called and one would be called every render, so I guess technically it wouldn't violate the rule, but I think the lint rule might flag it regardless, which will make others question it when they come across it.  

I'm also not sure how React deals with the dependency array changing length between renders.  It might be fine with it, but it also might do something weird.  I'd have to run a few experiments to check.

Assuming it all works,  I could see it becoming something like:

```js
const mapStateRef = useRef(mapState)
mapStateRef.current = mapState

const subspacer = typeof mapStateRef.current === "function"
  ? useMemo(() => subspace((...args) => mapStateRef.current(...args), namespace), [namespace])
  : useMemo(() => subspace(mapStateRef.current, namespace), [mapStateRef.current, namespace])
```

mpeyper: Note, the changing dependencies length does produce a warning out of React.  Changing the array to include `mapStateRef` satisfies this requirement.

travikk: huh, that's interesting;
my whole point was just mostly about - it's not immediately obvious what's happening there

mpeyper: Do you find the new changes easier?

travikk: Yeah.

packages/react-redux-subspace/src/hooks/useSubspace.js
```diff
@@ -6,31 +6,31 @@
...
 
-import { useContext, useMemo } from "react"
+import { useRef, useMemo, useContext } from "react"
 import { ReactReduxContext } from "react-redux"
 import { subspace } from "redux-subspace"
-import isObject from '../utils/isObject'
+import isObject from "../utils/isObject"
+import isFunction from "../utils/isFunction"
 
-export const useSubspaceAdvanced = (
```
jpeyper: Sorry, I know you don't want to, but this is a breaking change and this needs to be a major bump. semver can be inconvenient... but its there for a reason. The upgrade will be easy for most (all) dependents

jpeyper: I guess it begs the question, can we leave leave this function here and deprecate it instead of removing it?

packages/react-redux-subspace/src/hooks/useSubspace.js
```diff
@@ -6,31 +6,31 @@
...
-  return useSubspaceAdvanced(mapState, namespace, undefined, options)
+  const { context = ReactReduxContext } = options || {}
+
+  const mapStateFunc = useRef(mapState)
+  mapStateFunc.current = mapState
+
+  const subspacer = isFunction(mapState)
+    ? useMemo(
+        () => subspace((...args) => mapStateFunc.current(...args), namespace),
+        [mapStateFunc, namespace]
```
jpeyper: `mapStateFunc` is never going to change right? are you better off using something more obviously constant to pad out the dependencies array like `true` or (defined elsewhere) `REACT_NEEDS_DEPENDENCIES_ARRAY_TO_NOT_CHANGE_LENGTH`?

(or have a missed the point?)

packages/react-redux-subspace/src/hooks/useSubspace.js
mpeyper: This function was never exposed out if the index.js, so I'm not worried about removing it. The breaking change is the decorator prop that you could pass to `SubspaceProvider` that no longer exists that was the cause for this function to exist.

mpeyper: Note that that prop was never documented and was essentially an internal performance optimisation when using the HOC interface

packages/react-redux-subspace/src/hooks/useSubspace.js
mpeyper: You could do something like that. React doesn't care what the prop is, just that it doesn't change length.

I think I prefer the current solution anyway. While the result of `useRef` is static, our code has no guarantees of that, and the implementation of different renderers might be different. This way, if it did change, we would be covered.

jpeyper: thats fair.

packages/react-redux-subspace/src/hooks/useSubspace.js
jpeyper: I was sure i followed this out the index exports... guess its too early in the morning to be reviewing.

I'll leave it up to you, but for me, the public interface is changing... its a major bump.

mpeyper: Unfortunately I feel the same way, but I feels like the negative kind of churn for such a small break.

jpeyper: The upgrade path is negligible... I wouldn't worry too much about it. Anyone on v4/5 can just move to it easily, and anyone older than that probably has more work with the other dependencies to upgrade to meet out newer peer deps.

mpeyper: That's probably true.

mpeyper:
Thank you all for for your time and effort in this review.  I'm happy enough with this now and will go ahead and merge it.

@travikk, I've added a contribution for your review effort.  Please let me know if you don't want this and I'll remove the entry.

PR #149, opened 2019-09-24
Fix issue where wormholes create new state objects when the state has not changed

mpeyper:
| Q                       | A 
| Fixed Issues?           | 
| Documentation only PR   | 
| Patch: Bug Fix?         | 
| Minor: New Feature?     | 👍
| Major: Breaking Change? | 

In a similar vein to #137, this is an optimisation to reduce the chance of triggering unnecessary rerenders because things appear to be changing, when in reality, nothing is changing at all.

In this case, the current implementation of `redux-subspace-wormhole` will create a new state object each time `getState` is called, even if the state of the subspace and the mapped wormhole state (generally fairly static values) don't change.

Other than just being _more correct_ about how we deal with Redux's state object, the most common reason why `getState` would be and neither of these two values has changed is if another subspace dispatched an action and updated it's state, all the connected components will call `getState` to see if they need to rerender.

Generally, the cost of this is fairly low as the `mapStateToProps` would shallowly equal the previous render and the usual `connect` optimisations would kick in, but in cases where non-pure component are being used, or when `mapStateToProps` is particularly expensive, this could still pose an issue.

There's also more to Redux than just `react-redux` and we can't know that other parts of the ecosystem don't rely on the referential equality of `getState` when nothing changes, so we should be good citizens and not break that contract.

I guess it's worth mentioning that I did add an additional dependency to `redux-subspace-wormhole`, being `memoize-one`.  I felt the benefit of not having to write and maintain the memoize functionality myself was worth it for [the cost of that package](https://bundlephobia.com/result?p=memoize-one@5.1.1) and it's fairly common in the react/redux/modern js world that it's likely to already be in people's bundles anyway.

### DIFF
--- a/packages/redux-subspace-wormhole/src/wormhole.js
+++ b/packages/redux-subspace-wormhole/src/wormhole.js
@@ -6,30 +6,33 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import isPlainObject from 'lodash.isplainobject'
+import isPlainObject from "lodash.isplainobject"
+import memoizeOne from "memoize-one"
 
 const wormhole = (mapState, key) => {
+  if (typeof mapState === "string") {
+    const mapStateKey = mapState
+    mapState = state => state[mapStateKey]
 
-    if (typeof mapState === 'string') {
-        const mapStateKey = mapState
-        mapState = (state) => state[mapStateKey]
-
-        if (typeof key === 'undefined') {
-            key = mapStateKey
-        }
+    if (key === undefined) {
+      key = mapStateKey
     }
+  }
 
-    return (store) => ({
-        getState: (next) => () => {
-            const state = next()
+  return store => {
+    const createWormhole = memoizeOne((state, key, value) => ({ [key]: value, ...state }))
 
-            if (isPlainObject(state)) {
-                return { [key]: mapState(store.rootStore.getState()), ...state }
-            } else {
-                return state
-            }
+    return {
+      getState: next => () => {
+        const state = next()
+        if (isPlainObject(state)) {
+          return createWormhole(state, key, mapState(store.rootStore.getState()))
+        } else {
+          return state
         }
-    })
+      }
+    }
+  }
 }
 
 export default wormhole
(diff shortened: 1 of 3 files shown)


packages/redux-subspace-wormhole/src/wormhole.js
```diff
@@ -6,30 +6,33 @@
...
 
-            if (isPlainObject(state)) {
-                return { [key]: mapState(store.rootStore.getState()), ...state }
-            } else {
-                return state
-            }
+    return {
+      getState: next => () => {
+        const state = next()
+        if (isPlainObject(state)) {
```
jpeyper: I know this was here before, bit when is state not a plain object? Something like immutablejs?

Should the else case give some kind of warning that the store state is not compatible with wormholes?

mpeyper: A subspace could map to an array of primative value. In those cases, the state I'd still valid for that subspace, but we cannot apply a wormhole to it so we just don't.

mpeyper:
I'm starting to wonder if this is actually a patch, not a minor change?

packages/redux-subspace-wormhole/src/wormhole.js
jpeyper: Ok fair enough. Seems a bit invisible though.

Out of scope for this but perhaps we should do a one time Dev mode warning on the first encounter of a non object state just to warn the Dev they aren't getting wormholes?

jpeyper:
You said before, perf is a feature.

packages/redux-subspace-wormhole/src/wormhole.js
mpeyper: I think of rather put effort into completely reworking wormholes to be opt in rather than implicitly available. At that point warning makes more sense as we know the Dev cares. Currently we'd be making a grand assumption that every subspace uses the wormholes which I'm not sure we want to do.

mpeyper:
Don't use my words against me!

